PR #61, opened 2023-05-25
DPLT-996: state-indexer is too slow

kobayurii:
Disable the account_access_keys table logic and proxy the method


khorolets (changes requested):

rpc-server/src/modules/queries/methods.rs
```diff
@@ -222,6 +224,9 @@ pub async fn query(
             args,
         } => function_call(&data, block, account_id, &method_name, args.clone()).await,
         near_primitives::views::QueryRequest::ViewAccessKeyList { account_id } => {
+            #[cfg(not(feature = "account_access_keys"))]
+            return Ok(proxy_rpc_call(&data.near_rpc_client, params).await?);
```
khorolets, line 228: Here we just do a proxy if the feature is disabled (by default)

rpc-server/src/modules/queries/mod.rs
```diff
@@ -57,7 +57,10 @@ impl near_vm_logic::External for CodeStorage {
         ))
     }
 
-    #[cfg_attr(feature = "tracing-instrumentation", tracing::instrument(skip(self)))]
+    #[cfg_attr(
+        feature = "tracing-instrumentation",
+        tracing::instrument(skip(self, _mode))
```
khorolets, line 62: I had to skip `_mode` as well since it's not derives the Debug (IIRC the cargo check error)

state-indexer/src/configs.rs
```diff
@@ -543,6 +551,7 @@ impl ScyllaDBManager {
         Ok(result)
     }
 
+    #[cfg(feature = "account_access_keys")]
```
khorolets, line 554: I've disabled this function completely if the feature is disabled (by default). Also, I've used the full path to `HashMap` to avoid redundant feature-flag with the import at the top.

khorolets (approved):
I've addressed my review suggestions. I've left comments to highlight some of the changes.

PR #67, opened 2023-05-31
DPLT-1018: Introduce retries to the indexers writes to ScyllaDB

kobayurii:
This pr implement  a retry logic for the cases when we get an error from ScyllaDB and we cannot miss the data.

For indexers works in so-called “strict” mode when we retry infinitely though throwing a warning about it. In addition, we have a “non-strict” mode that limits the number of retries to N (5 by default).

For rpc-server works “non-strict” mode with limits the number of retries to 2.

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -72,6 +72,175 @@
 // ).await?,
 
 use scylla::prepared_statement::PreparedStatement;
+use scylla::retry_policy::{QueryInfo, RetryDecision};
+use scylla::transport::errors::QueryError;
+
+#[derive(Debug)]
+pub struct CustomDBRetryPolicy {
+    max_retry: u8,
+    strict_mode: bool,
+}
+
+impl CustomDBRetryPolicy {
+    pub fn new(max_retry: u8, strict_mode: bool) -> CustomDBRetryPolicy {
+        CustomDBRetryPolicy {
+            max_retry,
+            strict_mode,
+        }
+    }
+}
+
+impl Default for CustomDBRetryPolicy {
+    /// By default using `strict_mode`
+    fn default() -> CustomDBRetryPolicy {
+        CustomDBRetryPolicy::new(0, true)
+    }
+}
+
+impl scylla::retry_policy::RetryPolicy for CustomDBRetryPolicy {
+    /// Called for each new query, starts a session of deciding about retries
+    fn new_session(&self) -> Box<dyn scylla::retry_policy::RetrySession> {
+        Box::new(CustomRetrySession::new(self.max_retry, self.strict_mode))
+    }
+
+    /// Used to clone this RetryPolicy
+    fn clone_boxed(&self) -> Box<dyn scylla::retry_policy::RetryPolicy> {
+        Box::new(CustomDBRetryPolicy::new(self.max_retry, self.strict_mode))
+    }
+}
+
+struct CustomRetrySession {
+    max_retry: u8,
+    max_retry_count: u8,
+    strict_mode: bool,
+}
+
+impl CustomRetrySession {
+    pub fn new(max_retry: u8, strict_mode: bool) -> CustomRetrySession {
+        CustomRetrySession {
+            max_retry,
+            max_retry_count: 0,
+            strict_mode,
+        }
+    }
+
+    /// Decide if we should retry the query
+    pub fn should_retry(&mut self) -> bool {
+        // If `strict_mode` always retry
+        if self.strict_mode {
+            return true;
+        };
+        // If `max_retry_count` is more than `max_retry` don't retry
+        if self.max_retry_count > self.max_retry {
+            return false;
+        };
+        // If `max_retry_count` is less than `max_retry` increment `max_retry_count` and retry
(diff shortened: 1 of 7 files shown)


database/src/lib.rs
```diff
@@ -72,6 +72,159 @@
...
+            max_retry,
+            strict_mode,
+        }
+    }
+}
+
+impl Default for CustomRetryPolicy {
+    /// By default using 5 retries
+    fn default() -> CustomRetryPolicy {
+        CustomRetryPolicy::new(5, false)
```
khorolets, line 96: Strict mode should be on by default

database/src/lib.rs
```diff
@@ -72,6 +72,159 @@
 // ).await?,
 
 use scylla::prepared_statement::PreparedStatement;
+use scylla::retry_policy::{QueryInfo, RetryDecision};
+use scylla::transport::errors::QueryError;
+
+#[derive(Debug)]
+pub struct CustomRetryPolicy {
```
khorolets, line 79: Since it is related to the database I'd suggest renaming it to `CustomDBRetryPolicy` to be more explicit

database/src/lib.rs
```diff
@@ -72,6 +72,159 @@
...
+impl CustomRetrySession {
+    pub fn new(max_retry: u8, strict_mode: bool) -> CustomRetrySession {
+        CustomRetrySession {
+            max_retry,
+            max_retry_count: 0,
+            strict_mode,
+        }
+    }
+
+    pub fn is_retry(&mut self) -> bool {
```
khorolets, line 127: I guess `should_retry` would be a better name here.

khorolets: Doc-string explaining the method would be helpful too

database/src/lib.rs
```diff
@@ -72,6 +72,159 @@
...
+    }
+}
+
+impl scylla::retry_policy::RetrySession for CustomRetrySession {
+    /// Called after the query failed - decide what to do next
+    fn decide_should_retry(&mut self, query_info: QueryInfo) -> RetryDecision {
+        if let scylla::frame::types::LegacyConsistency::Serial(_) = query_info.consistency {
+            return RetryDecision::DontRetry;
+        };
+        tracing::warn!("Scylla RetryPolicy error: {:?}", query_info.error);
```
khorolets, line 155: The message clearly says that the error in RetryPolicy has occurred. But it is misleading. The error occurred in the query.

I suggest changing the message to clearly say about Database error.

Also, I see a huge `match` below dealing with different scenarios, perhaps it's worth adding logs there mentioning the info from the comments those match arms have for our own sake.

database/src/lib.rs
```diff
@@ -72,6 +72,159 @@
...
+            }
+            // Connection to the contacted node is overloaded, try another one
+            QueryError::UnableToAllocStreamId => RetryDecision::RetryNextNode(None),
+            // In all other cases propagate the error to the user
+            _ => RetryDecision::DontRetry,
+        }
+    }
+
+    /// Reset before using for a new query
+    fn reset(&mut self) {
```
khorolets, line 224: `reset_max_retry_counter`

database/src/lib.rs
```diff
@@ -159,9 +316,17 @@ pub trait ScyllaStorageManager {
         scylla_user: Option<&str>,
         scylla_password: Option<&str>,
         keepalive_interval: Option<u64>,
+        max_retry: u8,
+        strict_mode: bool,
     ) -> anyhow::Result<scylla::Session> {
-        let mut session: scylla::SessionBuilder =
-            scylla::SessionBuilder::new().known_node(scylla_url);
+        let handle = scylla::transport::ExecutionProfile::builder()
```
khorolets, line 322: maybe `scylla_execution_profile_handle`

khorolets (changes requested):
Overall looks great, thanks for the PR. Please, address the comments I've left.

database/src/lib.rs
kobayurii: This is method of `scylla::retry_policy::RetrySession` trait. We can't rename it.

database/src/lib.rs
```diff
@@ -72,6 +72,175 @@
...
+        if let scylla::frame::types::LegacyConsistency::Serial(_) = query_info.consistency {
+            return RetryDecision::DontRetry;
+        };
+        tracing::warn!("ScyllaDB QueryError: {:?}", query_info.error);
+        match query_info.error {
+            QueryError::IoError(_)
+            | QueryError::DbError(scylla::transport::errors::DbError::Overloaded, _)
+            | QueryError::DbError(scylla::transport::errors::DbError::ServerError, _)
+            | QueryError::DbError(scylla::transport::errors::DbError::TruncateError, _) => {
+                tracing::info!(
```
khorolets, line 162: I believe it should be `warn` but not `info` for sure. This applies to all logs in this method

khorolets (changes requested):
I believe it's the last changes iteration I am asking for and we're good to merge.

khorolets (approved):
LGTM!

PR #71, opened 2023-06-02
refactor: Upgrade near primitives crates, bump rust version, fix the code after the upgrade

khorolets:
With the release of `nearcore` 1.34.0, certain primitive crates were updated. In this pull PR, I addressed the breaking changes that were introduced.

The breaking changes primarily affected the `StoreKey` and `StoreValue`.

Additionally, the `near-jsonrpc-client-rs` crate is outdated. Since we heavily rely on it, I had to upgrade the primitives on the crate side and temporarily use my fork of it. We will need to modify the dependency once the new version of the crate is released.

On a positive note, starting from version `0.17.0` of the primitives, we are no longer limited to a specific old version of `tokio`. Therefore, I also updated the `tokio` version.


rpc-server/src/modules/blocks/utils.rs
```diff
@@ -278,6 +278,8 @@ pub(crate) fn is_matching_change(
             }
             | StateChangeValueView::DataDeletion { account_id, key } = &change.value
             {
+                let key: Vec<u8> = key.clone().into();
+                let key_prefix: Vec<u8> = key_prefix.clone().into();
```
khorolets, line 282: I'm having a bit of a bad times using the `StoreKey` and `StoreValue` after the `nearcore` commit https://github.com/near/nearcore/commit/416b38ce10f844b4f1936d18759d33a159d9f571

Perhaps, I am missing something, though I wonder why `DeriveAsRef` was replaced with `Deref`, couldn't we keep both?
cc @mina86 Would you have a moment to comment, please?

frol (approved):

rpc-server/src/modules/blocks/utils.rs
mina86: It just felt redundant.  Deref lets you access all of Vec’s methods and in a pinch you can do `&*key` to get reference to the vector.  I think those two lines should just be:

```suggestion
                let key: Vec<u8> = key.into();
                let key_prefix: Vec<u8> = key_prefix.into();
```

just like on line 121 in methods.rs file above.

khorolets: Thank you @mina86 for so quick response.

Unfortunately, it's not that simple. If I change the code according to your suggestion I get the error:

```
error[E0277]: the trait bound `Vec<u8>: std::convert::From<&StoreKey>` is not satisfied
   --> rpc-server/src/modules/blocks/utils.rs:281:40
    |
281 |                 let key: Vec<u8> = key.into();
    |                                        ^^^^ the trait `std::convert::From<&StoreKey>` is not implemented for `Vec<u8>`
    |
```

Trying to use `&*key` leads to:

```
error[E0277]: the trait bound `StoreKey: AsRef<[u8]>` is not satisfied
   --> rpc-server/src/modules/blocks/utils.rs:284:37
    |
284 |                     && hex::encode(&*key).starts_with(&hex::encode(key_prefix))
    |                        -----------  ^^^^ the trait `AsRef<[u8]>` is not implemented for `StoreKey`
    |                        |
    |                        required by a bound introduced by this call
    |
    = note: required for `&StoreKey` to implement `AsRef<[u8]>`
note: required by a bound in `hex::encode`
   --> /Users/khorolets/.cargo/registry/src/github.com-1ecc6299db9ec823/hex-0.4.3/src/lib.rs:259:18
    |
259 | pub fn encode<T: AsRef<[u8]>>(data: T) -> String {
    |                  ^^^^^^^^^^^ required by this bound in `encode`
help: consider dereferencing here
    |
284 |                     && hex::encode(&**key).starts_with(&hex::encode(key_prefix))
    |
```

**Compiler suggest to add more asterisks and it works while looks weird**

mina86: Ah, I see, `key` and `key_prefix` are references.  Here `hex::encode(key.as_slice())` or `hex::encode(&key[..])` should work.  `&**` works because first `&StoreKey` is dereferenced to `StoreKey` and then that is dereferenced to `Vec<u8>` which you then borrow.

Funnily, plain `hex::encode(key)` would work if `hex::encode` was written to accept `&[u8]` argument rather than `impl AsRef<[u8]>` argument.

Either way, there wasn’t any strong reason to remove `derive_more::AsRef` so adding it back is definitely an option as far as I’m concerned.

khorolets: Wow, TIL thanks for your help, indeed `.as_slice()` works like a charm. I am a bit embarrassed though :)

Again, thank you for being so responsive and for all your help and contributions to the ecosystem! ❤️

PR #74, opened 2023-06-07
DPLT-1029: Total method hits counters

kobayurii:
This pr implements  a set of metrics to accompany the unexpected proxy calls (metrics implemented in https://pagodaplatform.atlassian.net/browse/DPLT-969) so we can set up the Grafana in a more convenient way.


khorolets (approved):
Great! Thanks!

PR #76, opened 2023-06-12
refactor(rpc-server): Refactor shadow_compare_results function to return ShadowDataConsistencyError

khorolets:
Essentially, this PR does what the title says.

The goal was to refactor the `shadow_compare_results` function from the feature `shadow_data_consistency`. We wanted to change the way the function can be used as a result of this refactor.

Initially, the function `shadow_compare_results` were emitting log-messages. However, we find it inconvenient and wanted to be able to use this function a little bit differently.

That's why this PR introduces an error structure `ShadowDataConsistencyError` (hidden behind the same feature-flag)

```
/// Represents the error that can occur during the shadow data consistency check.
#[cfg(feature = "shadow_data_consistency")]
#[derive(thiserror::Error, Debug)]
pub enum ShadowDataConsistencyError {
    #[error("Failed to parse ReadRPC response: {0}")]
    ReadRpcResponseParseError(serde_json::Error),
    #[error("Failed to parse NEAR JSON RPC response: {0}")]
    NearRpcResponseParseError(serde_json::Error),
    #[error("NEAR RPC call error: {0}")]
    NearRpcCallError(String),
    #[error("Results don't match: {0}")]
    ResultsDontMatch(String),
}
```

This allows us to refactor the RPC methods. We are able to operate the comparison results and decide whether we need to proxy the call in case of an error on the Read RPC side, or we need to simply respond with the error crafted by Read RPC itself.

**Please note,** we are actively using the feature `shadow_data_consistency` to identify misalignments and bugs on the Read RPC side regarding the data consistency. After the bug-hunting cycle, this feature will not be used.

### DIFF
--- a/rpc-server/src/utils.rs
+++ b/rpc-server/src/utils.rs
@@ -94,23 +94,28 @@ pub async fn update_final_block_height_regularly(
 /// `readrpc_response`: a `Result<serde_json::Value, serde_json::Error>` object representing the results from Read RPC.
 /// `client`: `near_jsonrpc_client::JsonRpcClient`.
 /// `params`: `near_jsonrpc_client::methods::RpcMethod` trait.
+///
+/// In case of a successful comparison, the function returns `Ok(())`.
+/// Otherwise, it returns `Err(ShadowDataConsistencyError)`.
 #[cfg(feature = "shadow_data_consistency")]
 pub async fn shadow_compare_results<M>(
     readrpc_response: Result<serde_json::Value, serde_json::Error>,
     client: near_jsonrpc_client::JsonRpcClient,
     params: M,
-) where
+) -> Result<(), ShadowDataConsistencyError>
+where
     M: near_jsonrpc_client::methods::RpcMethod + std::fmt::Debug,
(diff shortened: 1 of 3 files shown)


kobayurii (approved):

rpc-server/src/utils.rs
```diff
@@ -94,37 +94,47 @@ pub async fn update_final_block_height_regularly(
...
 {
-    tracing::debug!("Compare results. {:?}", params);
+    tracing::debug!(target: "shadow_data_consistency", "Compare results. {:?}", params);
 
     let readrpc_response_json = match readrpc_response {
         Ok(readrpc_response_json) => readrpc_response_json,
         Err(err) => {
-            tracing::error!(target: "is_not_consistency", "Parse hundredx response error: {:#?}", err);
-            return;
+            // TODO: remove this log record after the methods calling function `shadow_compare_results` are refactored
```
khorolets, line 116: I've added the `TODO`s like this to smooth the transition since @kobayurii is going to refactor the methods and only after that, we can say we're fully adopted the refactored version.

PR #83, opened 2023-06-14
DPLT-1033: refactoring(rpc_server) `block` and `chunk` methods

kobayurii:
For now, we want to ensure Read RPC error is expected by doing a shadow call to the real RPC
This pr include refactoring next methods:
`block `
`chunks`
`EXPERIMENTAL_changes`
`EXPERIMENTAL_changes_in_block`

### DIFF
--- a/rpc-server/src/modules/blocks/methods.rs
+++ b/rpc-server/src/modules/blocks/methods.rs
@@ -58,6 +58,7 @@ pub async fn block(
             }
             Err(err) => {
                 tracing::warn!(target: "shadow_data_consistency", "Shadow data check: ERROR\n{}\n{:?}", error_meta, err);
+                crate::metrics::BLOCK_PROXIES_TOTAL.inc()
             }
         };
         block_response
@@ -91,6 +92,7 @@ pub async fn chunk(
             }
             Err(err) => {
                 tracing::warn!(target: "shadow_data_consistency", "Shadow data check: ERROR\n{}\n{:?}", error_meta, err);
+                crate::metrics::CHUNK_PROXIES_TOTAL.inc()
             }
         }
     }
@@ -138,6 +140,7 @@ pub async fn changes_in_block(
             }
             Err(err) => {
                 tracing::warn!(target: "shadow_data_consistency", "Shadow data check: ERROR\n{}\n{:?}", error_meta, err);
+                crate::metrics::CHNGES_IN_BLOCK_PROXIES_TOTAL.inc()
             }
         }
     }
@@ -187,6 +190,7 @@ pub async fn changes_in_block_by_type(
             }
             Err(err) => {
                 tracing::warn!(target: "shadow_data_consistency", "Shadow data check: ERROR\n{}\n{:?}", error_meta, err);
+                crate::metrics::CHNGES_IN_BLOCK_BY_TYPE_PROXIES_TOTAL.inc()
             }
         }
     }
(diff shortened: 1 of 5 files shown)


rpc-server/src/modules/blocks/methods.rs
```diff
@@ -1,51 +1,249 @@
...
+    }
+    crate::metrics::BLOCK_REQUESTS_TOTAL.inc();
+
+    #[cfg(not(feature = "shadow_data_consistency"))]
+    let result = fetch_block(&data, params.block_reference.clone()).await;
+
+    #[cfg(feature = "shadow_data_consistency")]
+    let result = {
+        /// For the shadow comparison we need to be sure that the block height is the same for all requests
+        let block = fetch_block_from_cache_or_get(&data, params.block_reference.clone()).await;
```
khorolets, line 38: I believe you can use the `result` you've got for the feature disabled. Instead of using here `params.block_reference` you can get the block by its height.

Furthermore, you can even reuse the `block_height` from the original `result` for the shadow call, can you?

What do think? Am I missing something?

rpc-server/src/modules/blocks/mod.rs
```diff
@@ -7,4 +7,5 @@ pub struct CacheBlock {
     pub block_height: near_primitives::types::BlockHeight,
     pub block_timestamp: u64,
     pub latest_protocol_version: near_primitives::types::ProtocolVersion,
+    pub chunks_count: u64,
```
khorolets, line 10: in `nearcore` the similar field is named `chunks_included`, please use that name instead

khorolets (changes requested):
Overall looks great! Though, a few changes are needed here and there before we can merge it.

rpc-server/src/modules/blocks/methods.rs
```diff
@@ -1,51 +1,253 @@
...
+    // Increase the OPTIMISTIC_REQUESTS_TOTAL metric if the request has optimistic finality.
+    if let near_primitives::types::BlockReference::Finality(finality) = &params.block_reference {
+        // Finality::None stands for optimistic finality.
+        if finality == &near_primitives::types::Finality::None {
+            crate::metrics::OPTIMISTIC_REQUESTS_TOTAL.inc();
+        }
+    }
+    crate::metrics::BLOCK_REQUESTS_TOTAL.inc();
+
+    #[cfg(not(feature = "shadow_data_consistency"))]
```
khorolets, line 32: Since we decided that we're going to respond with the response Read RPC has crafted, I'd want to get rid of redundant `#[cfg(not(feature = "shadow_data_consistency"))]` here.

You just don't need to put all the logic of the feature into the `let result = { .. }`, but just do it in the `{ .. }`. Our goal is to print a log with the shadow data check result. We don't have to return any results from there. Please, see the `query` for a reference.

khorolets: And you did it as I am asking in the `chunk` method below.

rpc-server/src/modules/blocks/methods.rs
```diff
@@ -1,51 +1,253 @@
...
         },
         near_primitives::types::BlockReference::Finality(finality) => match finality {
-            near_primitives::types::Finality::Final => data
+            near_primitives::types::Finality::Final => Ok(data
                 .final_block_height
-                .load(std::sync::atomic::Ordering::SeqCst),
-            _ => anyhow::bail!("Finality other than final is not supported"),
+                .load(std::sync::atomic::Ordering::SeqCst)),
+            _ => Err(
+                near_jsonrpc_primitives::types::blocks::RpcBlockError::UnknownBlock {
```
khorolets, line 227: In our case, it's `InternalError` I think. This applies to the below cases as well.

khorolets (changes requested):
Good. Just a few more things to correct, please.

khorolets (approved):
Great! Thank you!

PR #87, opened 2023-06-19
DPLT-1038: EXPERIMENTAL_tx_status receipts order

kobayurii:
We get a lot of errors about not matching when using the `shadow_data_consistency` feature.  
This often happens through different ordering of keys in the object. 
The basic idea is to sort the json responses before comparing them. 

This PR implements 2 functions for json sorting. 
The first `json_sort_value` sorts the keys using BTreeMap. 
The second `generate_array_key` is an additional function for sorting arrays. 
The idea is to sort the elements in the array by the values of the keys of the objects in the array.

I think it affects the performance of the service, but it should help us understand if our service is working correctly

### DIFF
--- a/rpc-server/src/utils.rs
+++ b/rpc-server/src/utils.rs
@@ -170,9 +170,11 @@ where
 
     let config = Config::new(CompareMode::Strict).numeric_mode(NumericMode::AssumeFloat);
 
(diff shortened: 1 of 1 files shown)


khorolets (approved):
LGTM!

PR #90, opened 2023-06-20
DPLT-1039: `query.view_code`

kobayurii:
Improvement  query view_code error handling

NO_CONTRACT_CODE when there is an account but no contract code
UNKNOWN_ACCOUNT when there is no contract code AND no account


khorolets (approved):
LGTM! Appreciate your improvements on the `ScyllaDbManager` side 👍

PR #91, opened 2023-06-27
DPLT-1045: Remove optimistic queries from error counters

kobayurii:
There is an assumption that optimistic queries we don’t support in the Read RPC (yet) cause most errors. 
This PR remove optimistic requests from the error counters.
Added SYNC_CHECKPOINT_REQUESTS_TOTAL counter.

### DIFF
--- a/rpc-server/src/modules/blocks/methods.rs
+++ b/rpc-server/src/modules/blocks/methods.rs
@@ -9,24 +9,162 @@ use crate::modules::blocks::utils::{
 use crate::utils::shadow_compare_results;
 use jsonrpc_v2::{Data, Params};
 
+use crate::utils::proxy_rpc_call;
 use near_primitives::trie_key::TrieKey;
 use near_primitives::views::StateChangeValueView;
 
+/// `block` rpc method implementation
+/// calls proxy_rpc_call to get `block` from near-rpc if request parameters not supported by read-rpc
+/// as example: block_id by SyncCheckpoint is not supported by read-rpc
+/// another way to get `block` from read-rpc using `block_call`
 #[allow(unused_mut)]
 #[cfg_attr(feature = "tracing-instrumentation", tracing::instrument(skip(data)))]
 pub async fn block(
     data: Data<ServerContext>,
     Params(mut params): Params<near_jsonrpc_primitives::types::blocks::RpcBlockRequest>,
 ) -> Result<near_jsonrpc_primitives::types::blocks::RpcBlockResponse, RPCError> {
-    tracing::debug!("`block` called with parameters: {:?}", params);
-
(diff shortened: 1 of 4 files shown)


rpc-server/src/modules/blocks/methods.rs
```diff
@@ -19,48 +20,94 @@ pub async fn block(
...
-            Ok(res) => {
-                if let near_primitives::types::BlockReference::Finality(_) = params.block_reference
+    match &params.block_reference {
+        near_primitives::types::BlockReference::SyncCheckpoint(_) => {
+            crate::metrics::SYNC_CHECKPOINT_REQUESTS_TOTAL.inc();
+            let block_view = proxy_rpc_call(&data.near_rpc_client, params).await?;
+            Ok(near_jsonrpc_primitives::types::blocks::RpcBlockResponse { block_view })
+        }
+        near_primitives::types::BlockReference::Finality(finality) => {
+            if finality != &near_primitives::types::Finality::Final {
```
khorolets, line 31: I think it'd be more honest to compare `finality` with `None` here

kobayurii: We supports only Finality::Final. In addition to Finality::None we have an additional case of Finality::DoomSlug, which should also be proxied

rpc-server/src/modules/blocks/methods.rs
```diff
@@ -18,15 +19,34 @@ pub async fn block(
...
+        }
+        near_primitives::types::BlockReference::BlockId(_) => {
+            block_call(data, Params(params)).await
         }
     }
+}
+
+#[allow(unused_mut)]
+#[cfg_attr(feature = "tracing-instrumentation", tracing::instrument(skip(data)))]
+pub async fn block_call(
```
khorolets, line 45: I guess it doesn't have to be public, does it?

rpc-server/src/modules/blocks/methods.rs
```diff
@@ -104,13 +151,34 @@ pub async fn changes_in_block(
...
+        }
+        near_primitives::types::BlockReference::BlockId(_) => {
+            changes_in_block_call(data, Params(params)).await
         }
     }
+}
+
+#[allow(unused_mut)]
+#[cfg_attr(feature = "tracing-instrumentation", tracing::instrument(skip(data)))]
+pub async fn changes_in_block_call(
```
khorolets, line 175: Do you know if this is required to be public?

khorolets (changes requested):
Please, check the new functions (`*_call`) you've added. Do they require to be public?

Also, it'd be really helpful if you added doc-strings to the functions so it is easier to understand what are they for. Could you do this, please?

Issue #92, opened 2023-07-13
`shadow_data_consistency` logic does not work


telezhnaya:
I've checked 2 "broken" blocks, 96382302 and 34955619
Both of them are equal if we ignore the order of the fields in dicts (the order of the fields in lists is correct)

I decided to dive into `shadow_data_consistency` logic.

I see there is an attempt to sort the fields in the function `json_sort_value`
But after sorting, we put values back into `serde_json::Value::Object` which does not hold the order by default. So this function does not do anything.

We can try to play with `preserve_order` feature for `serde_json::Value::Object`, it should make the result stable, but I'd personally prefer not to do this manually and just use the library for comparing jsons instead: https://docs.rs/assert-json-diff/latest/assert_json_diff/

PR #93, opened 2023-07-20
Improvement(rpc-server): Add additional logs for `shadow_data_consistency`

kobayurii:
https://github.com/near/read-rpc/issues/92
In this pr we improve shadow_data_consistency logs. 

Data may mismatch in 4 cases:
1. Both services(read_rpc and near_rpc) have a successful response 
if data mismatch we are logging into `shadow_data_consistency_successful_responses`both response objects for future investigation.
2. read_rpc - failure, near_rpc - successful
we are logging into `shadow_data_consistency_read_rpc_failure_response` just message about mismatch data without objects. In the future we should be investigate why we got error in the read_rpc.
3. read_rpc - successful , near_rpc - failure
we are logging into `shadow_data_consistency_near_failure_response` just message about mismatch data without objects. Expected that all error will be related with network issues.
4. Both services have a failure response
we are logging into `shadow_data_consistency_failure_responses` both response objects for future investigation.
Expected we will only have a difference in the error text.

### DIFF
--- a/rpc-server/src/modules/queries/methods.rs
+++ b/rpc-server/src/modules/queries/methods.rs
@@ -116,7 +116,7 @@ async fn query_call(
             )
         }
 
-        let (read_rpc_response_json, response_success) = match &result {
+        let (read_rpc_response_json, is_response_ok) = match &result {
             Ok(res) => (serde_json::to_value(res), true),
             Err(err) => (serde_json::to_value(err), false),
         };
@@ -125,7 +125,7 @@ async fn query_call(
             read_rpc_response_json,
             near_rpc_client,
             params,
-            response_success,
+            is_response_ok,
         )
         .await;
 
--- a/rpc-server/src/modules/queries/methods.rs
+++ b/rpc-server/src/modules/queries/methods.rs
@@ -116,13 +116,18 @@ async fn query_call(
             )
         }
 
-        let read_rpc_response_json = match &result {
-            Ok(res) => serde_json::to_value(res),
-            Err(err) => serde_json::to_value(err),
+        let (read_rpc_response_json, response_success) = match &result {
+            Ok(res) => (serde_json::to_value(res), true),
+            Err(err) => (serde_json::to_value(err), false),
         };
 
-        let comparison_result =
-            shadow_compare_results(read_rpc_response_json, near_rpc_client, params).await;
+        let comparison_result = shadow_compare_results(
+            read_rpc_response_json,
+            near_rpc_client,
+            params,
+            response_success,
+        )
+        .await;
 
         match comparison_result {
             Ok(_) => {
--- a/rpc-server/src/modules/transactions/methods.rs
+++ b/rpc-server/src/modules/transactions/methods.rs
@@ -29,10 +29,9 @@ pub async fn tx(
     {
         let near_rpc_client = data.near_rpc_client.clone();
         let error_meta = format!("TX: {:?}", params);
-
-        let read_rpc_response_json = match &result {
-            Ok(res) => serde_json::to_value(res),
-            Err(err) => serde_json::to_value(err),
+        let (read_rpc_response_json, response_success) = match &result {
+            Ok(res) => (serde_json::to_value(res), true),
+            Err(err) => (serde_json::to_value(err), false),
         };
 
         let comparison_result = shadow_compare_results(
@@ -44,12 +43,13 @@ pub async fn tx(
             near_jsonrpc_client::methods::tx::RpcTransactionStatusRequest {
                 transaction_info: tx_status_request.transaction_info,
             },
+            response_success,
         )
         .await;
 
         match comparison_result {
             Ok(_) => {
-                tracing::info!(target: "shadow-data-consistency", "Shadow data check: CORRECT\n{}", error_meta);
+                tracing::info!(target: "shadow_data_consistency", "Shadow data check: CORRECT\n{}", error_meta);
             }
             Err(err) => {
                 tracing::warn!(target: "shadow_data_consistency", "Shadow data check: ERROR\n{}\n{:?}", error_meta, err);
@@ -79,9 +79,9 @@ pub async fn tx_status(
         let near_rpc_client = data.near_rpc_client.clone();
         let error_meta = format!("EXPERIMENTAL_TX_STATUS: {:?}", params);
 
-        let read_rpc_response_json = match &result {
-            Ok(res) => serde_json::to_value(res),
-            Err(err) => serde_json::to_value(err),
+        let (read_rpc_response_json, response_success) = match &result {
+            Ok(res) => (serde_json::to_value(res), true),
+            Err(err) => (serde_json::to_value(err), false),
         };
 
         let comparison_result = shadow_compare_results(
@@ -93,12 +93,13 @@ pub async fn tx_status(
             near_jsonrpc_client::methods::EXPERIMENTAL_tx_status::RpcTransactionStatusRequest {
                 transaction_info: tx_status_request.transaction_info,
             },
+            response_success,
         )
         .await;
 
         match comparison_result {
             Ok(_) => {
-                tracing::info!(target: "shadow-data-consistency", "Shadow data check: CORRECT\n{}", error_meta);
+                tracing::info!(target: "shadow_data_consistency", "Shadow data check: CORRECT\n{}", error_meta);
             }
             Err(err) => {
                 tracing::warn!(target: "shadow_data_consistency", "Shadow data check: ERROR\n{}\n{:?}", error_meta, err);
--- a/rpc-server/src/modules/transactions/methods.rs
+++ b/rpc-server/src/modules/transactions/methods.rs
@@ -29,7 +29,7 @@ pub async fn tx(
     {
         let near_rpc_client = data.near_rpc_client.clone();
         let error_meta = format!("TX: {:?}", params);
-        let (read_rpc_response_json, response_success) = match &result {
+        let (read_rpc_response_json, is_response_ok) = match &result {
             Ok(res) => (serde_json::to_value(res), true),
             Err(err) => (serde_json::to_value(err), false),
         };
@@ -43,7 +43,7 @@ pub async fn tx(
             near_jsonrpc_client::methods::tx::RpcTransactionStatusRequest {
                 transaction_info: tx_status_request.transaction_info,
             },
-            response_success,
+            is_response_ok,
         )
         .await;
 
@@ -79,7 +79,7 @@ pub async fn tx_status(
         let near_rpc_client = data.near_rpc_client.clone();
         let error_meta = format!("EXPERIMENTAL_TX_STATUS: {:?}", params);
 
-        let (read_rpc_response_json, response_success) = match &result {
+        let (read_rpc_response_json, is_response_ok) = match &result {
             Ok(res) => (serde_json::to_value(res), true),
             Err(err) => (serde_json::to_value(err), false),
         };
@@ -93,7 +93,7 @@ pub async fn tx_status(
             near_jsonrpc_client::methods::EXPERIMENTAL_tx_status::RpcTransactionStatusRequest {
                 transaction_info: tx_status_request.transaction_info,
             },
-            response_success,
+            is_response_ok,
         )
         .await;
 
--- a/rpc-server/src/utils.rs
+++ b/rpc-server/src/utils.rs
@@ -103,7 +103,7 @@ pub async fn shadow_compare_results<M>(
     read_rpc_response: Result<serde_json::Value, serde_json::Error>,
     client: near_jsonrpc_client::JsonRpcClient,
     params: M,
-    read_rpc_response_success: bool,
+    read_rpc_response_is_ok: bool,
 ) -> Result<(), ShadowDataConsistencyError>
 where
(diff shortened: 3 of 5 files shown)


rpc-server/src/modules/blocks/utils.rs
```diff
@@ -6,6 +6,47 @@ use anyhow::Context;
 use near_primitives::views::{StateChangeValueView, StateChangesRequestView};
 use num_traits::ToPrimitive;
 
+async fn fetch_object_from_s3(
+    s3_client: &aws_sdk_s3::Client,
+    s3_bucket_name: &str,
+    key: String,
+) -> anyhow::Result<aws_sdk_s3::output::GetObjectOutput> {
+    let mut s3_response = s3_client
```
khorolets, line 14: This doesn't seem idiomatic to me.

Can you consider refactoring it like the following:

```rust
let mut retry_count: u8 = 0;

let s3_response = loop {
    match s3_client
        .get_object()
        .bucket(s3_bucket_name)
        .key(&key)
        .send()
        .await
    {
        Ok(response) => break Ok(response),
        Err(err) => match err {
            // TODO: add a comment explaining why retry on these errors
            aws_sdk_s3::types::SdkError::ConstructionFailure(_)
            | aws_sdk_s3::types::SdkError::TimeoutError(_)
            | aws_sdk_s3::types::SdkError::DispatchFailure(_)
            | aws_sdk_s3::types::SdkError::ResponseError(_) => {
                tracing::debug!("Error to get object from s3, Retry");
                if retry_count < 3 {
                    retry_count += 1;
                    continue;
                } else {
                    break Err(err);
                }
            }
            // TODO: add a comment why don't we retry on other errors
            _ => break Err(err),
        },
    }
};
```

Also, since we're passing the `key` already, we expect to know what kind of object we're fetching when we call the function. How about adding a generic to the function so we can pass the `*View` type for the fetched object to be parsed to?

For instance:

```rust
let block = fetch_object_from_s3::<BlockView>(&client, &bucket_name, format!("{:0>12}/block.json", block_height))?;
```

khorolets (changes requested):
I'd like you to add docstrings to the functions you introduce for readability. This is a general request from me.

Also, could you please consider the changes I've proposed? The change would also affect other places, so I'm not commenting on those.

I can't entirely agree with the place for the log message for data consistency. However, I don't have any better suggestions.

telezhnaya (commented):
Have you covered the thing that we should give different logs:
- when both answers are successful, but they do not match
- when one or both answers were not able to be collected properly
?

kobayurii:
> Have you covered the thing that we should give different logs:
> 
> * when both answers are successful, but they do not match
> * when one or both answers were not able to be collected properly
>   ?

I haven't forgotten.  I analyzed the situation and found a case when an error is normal. For example, if the block does not exist then both services return an error and we need to compare these errors and make sure that we return the same error as near_rpc. At the moment I just added the compared objects to the logging, I hope this will improve the situation with the investigation of mismatching.

khorolets (approved):
My requests and concerns were addressed. I am approving, though I wouldn't merge until @telezhnaya approves.

telezhnaya (changes requested):
Sorry, but this does not address the thing I've asked for. The initial idea was to simplify the logs and help to identify the cases when both services give successful results, and these results do not match. You haven't covered this.

kobayurii:
> Sorry, but this does not address the thing I've asked for. The initial idea was to simplify the logs and help to identify the cases when both services give successful results, and these results do not match. You haven't covered this.

You are right! I lost some moments and I will improve code coming soon. Sorry for that!

rpc-server/src/modules/transactions/methods.rs
```diff
@@ -93,6 +101,7 @@ pub async fn tx_status(
             near_jsonrpc_client::methods::EXPERIMENTAL_tx_status::RpcTransactionStatusRequest {
                 transaction_info: tx_status_request.transaction_info,
             },
+            response_state,
```
khorolets, line 104: I cannot comment on line 110 because it's not a part of the changes you make, but I've noticed there is a target `shadow-data-consistency` while on 113 the target is `shadow_data_consistency` (note the delimiters), please could you check while you're here to fix them?

khorolets (changes requested):
Thank you for the detailed PR description. I'm leaving some changes request, but I still need to review the PR. 
I am thinking about the way `ResponseState` works. I feel it should be improved, but I can't concentrate right now to tell how.

rpc-server/src/utils.rs
```diff
@@ -194,6 +223,15 @@ pub enum ShadowDataConsistencyError {
     ResultsDontMatch(String),
 }
 
+/// Represents the rpc response status for data comparison.
+#[cfg(feature = "shadow_data_consistency")]
+#[derive(Debug, Clone)]
+pub enum ResponseState {
```
telezhnaya, line 229: Here, you mix the results of 2 responses in one object. The name does not explain this.
Later, you pass this state to `shadow_compare_results` while we haven't even seen the result of Archival RPC, we will make this request later inside this function. It's overcomplicated.

I suggest having only 2 options here - success and failure.
As a result, with these 2 options, not sure we need a separate structure here at all.

rpc-server/src/utils.rs
```diff
@@ -171,10 +175,35 @@ where
...
+                tracing::warn!(
+                    target: "shadow_data_consistency_successful_response",
+                    "Shadow data check: DATA MISMATCH\n READ_RPC_DATA: {:?}\n NEAR_RPC_DATA: {:?}\n",
+                    read_rpc_json,
+                    near_rpc_json,
+                );
+            }
+            ResponseState::Failure => {
+                tracing::warn!(target: "shadow_data_consistency_failure_response",
+                    "Shadow data check: DATA MISMATCH\n READ_RPC_STATE: {:?}\n",
```
telezhnaya, line 195: This and the next log message does not explain the situation we have here (who failed, who succeeded)

rpc-server/src/modules/blocks/utils.rs
```diff
@@ -2,10 +2,54 @@ use crate::config::ServerContext;
 use crate::modules::blocks::methods::fetch_block;
 use crate::modules::blocks::CacheBlock;
 use crate::storage::ScyllaDBManager;
-use anyhow::Context;
 use near_primitives::views::{StateChangeValueView, StateChangesRequestView};
 use num_traits::ToPrimitive;
 
+/// Fetch object from s3 with retry
+/// Try to get 3 times if we have some problems with network
+async fn fetch_object_from_s3_with_retry<T>(
```
telezhnaya, line 10: note for the future: it's much better to open separate PRs for separate issues. This makes the merging faster (faster dopamine + 2 doses of it because we will merge 2 PRs instead of 1)
And it's much easier for the reviewers to look at smaller PRs

(I'm not suggesting splitting this exact PR, let's finish it as it is)

telezhnaya (changes requested):

rpc-server/src/modules/blocks/utils.rs
```diff
@@ -2,10 +2,54 @@ use crate::config::ServerContext;
...
+    s3_client: &aws_sdk_s3::Client,
+    s3_bucket_name: &str,
+    key: String,
+) -> anyhow::Result<T>
+where
+    T: serde::de::DeserializeOwned,
+{
+    let mut retry_count: u8 = 0;
+    let s3_response = loop {
```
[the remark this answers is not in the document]
kobayurii: In this case, I followed @khorolets instructions. I'm not sure I should go back to my old decision without consulting him.

rpc-server/src/modules/blocks/utils.rs
```diff
@@ -17,24 +61,15 @@ pub async fn fetch_block_from_s3(
...
-        .get_object()
-        .bucket(s3_bucket_name)
-        .key(format!("{:0>12}/block.json", block_height))
-        .send()
-        .await
+
+    match fetch_object_from_s3_with_retry::<near_primitives::views::BlockView>(
+        s3_client,
+        s3_bucket_name,
+        format!("{:0>12}/block.json", block_height),
```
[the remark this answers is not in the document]
kobayurii: I left a docstring for this.

rpc-server/src/modules/blocks/utils.rs
```diff
@@ -54,22 +89,13 @@ pub async fn fetch_shard_from_s3(
...
-        .await
-        .with_context(|| "Invalid data from s3")?
-        .into_bytes();
 
-    serde_json::from_slice::<near_indexer_primitives::IndexerShard>(body_bytes.as_ref())
-        .with_context(|| "Invalid serialised data")
+    fetch_object_from_s3_with_retry::<near_indexer_primitives::IndexerShard>(
+        s3_client,
+        s3_bucket_name,
+        format!("{:0>12}/shard_{shard_id}.json", block_height),
```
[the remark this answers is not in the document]
kobayurii: I don't think we need it. I don't think the lake framework is going to change the storage structure anytime soon.

rpc-server/src/modules/blocks/utils.rs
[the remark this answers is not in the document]
khorolets: This formats the number in a string of length 12 with leading zeroes

rpc-server/src/modules/blocks/utils.rs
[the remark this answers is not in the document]
khorolets: Loop is idiomatic here. Since this code is hidden behind a non-production feature it won't harm. Please keep it

rpc-server/src/modules/blocks/utils.rs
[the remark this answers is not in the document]
telezhnaya: So, every developer should:
1. remember this `{:0>12}`
2. check if someone does not make a typo (`{:0>13}`)
3. go and read a doc each time

Why, if we can just have a self-documented constant

rpc-server/src/modules/blocks/utils.rs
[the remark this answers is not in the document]
telezhnaya: It's only one argument, there are many others
https://github.com/near/read-rpc/pull/93#discussion_r1273437747

telezhnaya (approved):
I see no more major issues here.
But, this does not fix #92, please check it's not closed automatically after the merge
Thank you

rpc-server/src/utils.rs
```diff
@@ -171,10 +171,48 @@ where
...
 
     if let Err(err) = assert_json_matches_no_panic(&read_rpc_json, &near_rpc_json, config) {
+        // separate mismatching successful and failure responses into different targets
+        // to make it easier to find reasons of the mismatching
+
+        if read_rpc_response_success && near_rpc_response_success {
+            // Both services(read_rpc and near_rpc) have a successful response
+            // if data mismatch we are logging into `shadow_data_consistency_successful_responses`
+            // both response objects for future investigation.
+            tracing::warn!(target: "shadow_data_consistency_successful_responses",
```
khorolets, line 185: Introduction of couple of new targets really bothers me. Can you reuse the `shadow_data_consistency` we already have and instead define the cause of the warning in the log message itself, please?

rpc-server/src/modules/queries/methods.rs
```diff
@@ -116,13 +116,18 @@ async fn query_call(
             )
         }
 
-        let read_rpc_response_json = match &result {
-            Ok(res) => serde_json::to_value(res),
-            Err(err) => serde_json::to_value(err),
+        let (read_rpc_response_json, response_success) = match &result {
```
khorolets, line 119: I don't feel `response_success` is a good name for this variable. Maybe something like `is_response_ok` (`is_` could help identify it's a `bool`, `_ok` is closer to what we deal with). What do you guys think @kobayurii @telezhnaya ?

khorolets (changes requested):

rpc-server/src/modules/queries/methods.rs
telezhnaya: 100% agree

rpc-server/src/utils.rs
```diff
@@ -171,11 +171,66 @@ where
...
+                "shadow_data_consistency_failure_responses",
+                format!(
+                    "Shadow data check: DATA MISMATCH\n READ_RPC_DATA: {:?}\n NEAR_RPC_DATA: {:?}\n",
+                    read_rpc_json,
+                    near_rpc_json,
+                ),
+            )
+        };
+        return Err(ShadowDataConsistencyError::ResultsDontMatch(format!(
+            "{err}. Reason: {reason}. Description {description}"
```
khorolets, line 232: Can you post an example of such input for clarity, please?

kobayurii: [exampl.log](https://github.com/near/read-rpc/files/12237671/exampl.log)

khorolets:
@kobayurii @telezhnaya 
I've talked to @gmilescu and he suggested us splitting the errors in the counters. Something like:
- `block_error_proxies_counter_0` - RR success, AR success - data mismatch
- `block_error_proxies_counter_1` - RR success, AR error
- `block_error_proxies_counter_3` - RR error, AR success

This is pretty similar to what @kobayurii did in this PR for separating logs. What do you think? I guess we should merge this PR and open a separate followup issue to cover the idea described.

khorolets (approved):

PR #94, opened 2023-07-26
Improvement(rpc-server): Add retrying to fetch object from s3

kobayurii:
In this pr we added retrying to fetch object from s3. We add new function fetch_object_from_s3_with_retry in which we try to get object from s3 three times if we have some problem with network, otherwise we return an error.


khorolets (approved):
LGTM!

PR #95, opened 2023-07-27
refactor(tx-indexer): `tx-indexer` optimization (performance)

khorolets:
In this PR, I aim to optimize `tx-indexer` to improve performance and speed by doing small yet meaningful changes.

* 🟢 refactor(tx-indexer): Execute `index_transactions` and `update_meta` in parallel

I've noticed in the profiler that these two functions are executed sequentially in the `handle_streamer_message` while they can be joined.

* 🔴 REVERTED (no effect) refactor: Upgrade `scylla` to the latest, disable write coalescing to try if it improves the speed

`.write_coalescing(false)` parameter was added to the latest version of `scylla` crate. THe doc around it says:
> However, this optimization may worsen latency if the rate of requests issued by the application is low, but otherwise the application is heavily loaded with other tasks on the same tokio executor. Please do performance measurements before committing to disabling this option.

So I want to do measurements.

The next day, `scylla` was updated to a major `0.9.0` version.

* refactor(tx-indexer): Upgrade opentelementry crates, change the jaeger-tracing configuration to the CollectorPipeline

We've been using a bit old version of opentelemetry crates around Jaeger, so I've updated them to be sure we collect the tracing data properly when we do.

* 🟢 (gives ~30% performance improvement) refactor(database): Refactor prepare_statement function to take optional `Consistency`

The idea is to try to use `Consistency::Any` for the queries in `tx-indexer` since in the ScyllaDB docs it says:

> Closest replica, as determined by the snitch. If all replica nodes are down, write succeeds after a [hinted handoff](https://opensource.docs.scylladb.com/stable/glossary.html#term-Hinted-Handoff). Provides low latency, guarantees writes never fail.

However, the writes still take ~150ms even with `Consistency::Any` (before that we were using `Consistency::LocalQuorum`)

* 🟢⤵️ (dramatic improvement!) Introduce `SCYLLA_DC_HOST_FILTER` optional parameter to filter the Scylla connection to a single DC

@andrei-near noticed that, occasionally, writes to the ScyllaDB happen in less than 1ms while others are taking ~150ms. He shared his thoughts that the root cause of such behavior might lay in the way `scylla` is choosing what node to connect to. 
@andrei-near found a parameter available in `scylla`'s `Session` [`host_filter`](https://docs.rs/scylla/0.9.0/scylla/transport/session_builder/struct.GenericSessionBuilder.html#method.host_filter). Later he shared his ad-hoc experiment where he used the `host_filter` to filter to a single node in the same region and the BPS was ~200!. 
This commit enables a DC-based filter to be provided to filter the host to a single DC (in the same region as the indexer). And it improves the speed of the indexer **~33 times**

* 🟢 (confirmed it's working as expected) refactor: Replace Scylla host_filter with Load Balancing Policy

This is a common sense improvement to the previous approach. We're replacing the usage of `host_filter` which will lead to a disconnection if the DC is unreachable. That's not the preferable behavior, so we're switching to the Load Balancing Policy instead. Thus we're giving the **preferable** DC to connect while allowing the connection to others in case of troubles.

The environmental variable has been changed to `SCYLLA_PREFERRED_DC`

* refactor(database): Introduce wrapper prepare_read_query and prepare_write_query. Add doc about tracing.

This commit adds `prepare_write_query` and `prepare_read_query` to the `database` crate, thus introducing wrappers. For now, we're using `Consistency::LocalQorum` for read queries and `Any` for write ones. @telezhnaya has concerns about `LocalQorum`, though. However, we were using `LocalQorum` at the recommendation of the Scylla team. I want to keep it as is for now. But it will be very easy to change later.

Also, I've documented a tracing-related environment variable.


andrei-near:
Because of our previous experiment with Consistency Any improving performance by 30%, we may try to combine it with host_filter. Writes are guaranteed with this consistency we don't really loose anything. What do you think?

khorolets:
> Because of our previous experiment with Consistency Any improving performance by 30%, we may try to combine it with host_filter. Writes are guaranteed with this consistency we don't really loose anything. What do you think?

The results I've shared in the description include the `consistency::Any` already. Thus I agree with you.

database/src/lib.rs
```diff
@@ -344,6 +354,14 @@ pub trait ScyllaStorageManager {
             .known_node(scylla_url)
             .default_execution_profile_handle(scylla_execution_profile_handle);
 
+        if let Some(scylla_dc_host_filter) = scylla_dc_host_filter {
+            session = session.host_filter(std::sync::Arc::new(
+                scylla::transport::host_filter::DcHostFilter::new(
```
andrei-near, line 359: Oh, so the problem is not with establishment of TCP sessions but with Scylla driver using ALL datacenters (including the ones across ocean)! This explains better 150ms write duration.

andrei-near (approved):

database/src/lib.rs
khorolets: Yep, your experiment and results led me to this conclusion. So I've implemented this and confirmed that was the root cause. Now it seems obvious.

andrei-near:
Digging further in scylla lib, saw preffered_datacenter Policy option https://docs.rs/scylla/0.9.0/scylla/transport/load_balancing/struct.DefaultPolicyBuilder.html#method.prefer_datacenter. 

We should test with this approach and if we get good performance, we should is it instead of host_filter. In case of DC issues, Scylla will automatically use the remote DCs

khorolets:
> Digging further in scylla lib, saw preffered_datacenter Policy option https://docs.rs/scylla/0.9.0/scylla/transport/load_balancing/struct.DefaultPolicyBuilder.html#method.prefer_datacenter.
> 
> We should test with this approach and if we get good performance, we should is it instead of host_filter. In case of DC issues, Scylla will automatically use the remote DCs

Sounds reasonable, I'll take a closer look tomorrow 👍

kobayurii (approved):

database/src/lib.rs
```diff
@@ -309,9 +311,16 @@ pub trait ScyllaStorageManager {
     async fn prepare_query(
         scylla_db_session: &std::sync::Arc<scylla::Session>,
         query_text: &str,
+        consistency: Option<scylla::frame::types::Consistency>,
```
telezhnaya, line 314: What does consistency mean for read queries?

telezhnaya: My picture now: we should always use `quorum` (actually, even `all`) for read queries because there's a chance that we don't have data on some of our instances. It will provide us with the errors.

We may always use `any` for write queries to speed up the solution. Though other consistency levels are fine, it will just slow down the writing process.

Correct me if I'm wrong.
But, even if we have another picture, my concern here is that we should keep this all in mind during developing new features and reviewing the code. I suggest making wrappers, if it's possible (`prepare_write_query`, `prepare_read_query`).

telezhnaya (commented):
Your PR comments are as incredible as usual, I love it
Thank you so much for this great work!

state-indexer/src/configs.rs
```diff
@@ -386,6 +395,7 @@ impl ScyllaStorageManager for ScyllaDBManager {
                 "INSERT INTO state_indexer.account_access_keys
                     (account_id, block_height, active_access_keys)
                     VALUES(?, ?, ?)",
+                None,
```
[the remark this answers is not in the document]
khorolets: Yeah, we want. This is a leftover since I was focused on `tx-indexer` and decided to use the default value for others. Thank you for bringing this up. I'll adjust.

telezhnaya (approved):

PR #100, opened 2023-08-11
refactor(rpc-server): Shadow data consistency call to ignore `receipts_outcome.outcome.metadata`

khorolets:
This PR naively removes the `metadata` field from the JSON responses being compared during the `shadow_data_consistency` check. 

This field is not backward compatible on the `nearcore` side, it is used mainly for debugging, and different `nearcore` nodes can respond with different data in that field. But it harms our shadow data checks, so we want to ignore this field completely.

### DIFF
--- a/rpc-server/src/utils.rs
+++ b/rpc-server/src/utils.rs
@@ -310,7 +310,13 @@ fn json_sort_value(value: serde_json::Value) -> serde_json::Value {
                 .collect();
             serde_json::Value::from(new_value)
         }
-        serde_json::Value::Object(obj) => {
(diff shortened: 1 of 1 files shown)


telezhnaya (approved):
We should have a doc mentioning all unsupported fields from JSON RPC, could you please start collecting such cases?

khorolets:
> We should have a doc mentioning all unsupported fields from JSON RPC, could you please start collecting such cases?

What do you mean? We're not dropping the support, we don't compare it, but it is included in the response. No matter what the data for that field we have.

PR #102, opened 2023-08-15
fix(state-indexer): Skipped, duplicate chunks problem solved

khorolets:
This PR addresses an issue where ReadRPC would return an `Unavailable chunk` error for certain chunks. The root cause stemmed from chunk producers skipping chunks, leading to chunk duplication. The discrepancy between how the `state-indexer` writes data and the NEAR Indexer Framework handles such cases led to overwriting the `block_height` for the duplicated chunk, pointing to an incorrect block.

### Key Changes:

* Redesigned the chunks table.
* Reintroduced the blocks table.
* Adjusted rpc-server for accurate reads.

**Next Steps**: Post-merge, we must drop the chunk table and restart the state-indexer from the genesis for data recollection. Connection enhancements done in #95 ensure a faster process.


state-indexer/src/configs.rs
```diff
@@ -298,12 +303,25 @@ impl ScyllaStorageManager for ScyllaDBManager {
...
+
+        scylla_db_session
+            .query(
+                "
+                CREATE TABLE IF NOT EXISTS chunks (
                     chunk_hash varchar,
+                    block_height varint,
                     shard_id varint,
-                    PRIMARY KEY (chunk_hash)
+                    height_included varint,
```
telezhnaya, line 323: What's the difference between block_height and height_included?

rpc-server/src/storage.rs
```diff
@@ -290,4 +298,45 @@ impl ScyllaDBManager {
...
+        let rows = Self::execute_prepared_query(
+            &self.scylla_session,
+            &self.get_height_included_and_shard_id_by_block_height,
+            (num_bigint::BigInt::from(block_height),),
+        )
+        .await?
+        .rows()?;
+
+        let height_included_and_shard_id = rows
+            .into_typed::<(num_bigint::BigInt, num_bigint::BigInt)>()
```
telezhnaya, line 316: The code would be much cleaner if we introduce custom type with from()/try_from() method

telezhnaya (changes requested):

state-indexer/src/configs.rs
telezhnaya: Ah, OK
If the chunk is duplicated, we will have
```
block_height height_included
10           10
10           11
10           12
```
Right?

khorolets: > Right?

Exactly!

telezhnaya: It's non-intuitive.
Let's rename the columns to help future us understand this.

khorolets: > height_included is actually included_in_block_height
> 
> block_height could be first_appeared_at_ block_height (don't like this, but I don't have any better ideas for now)

For me, table name + field name `chunks.height_included` is not worse than `chunks.included_in_block_height`. And looks obvious.

`first_appeared_at_block_height` that'd be misleading for your example 

first_appeared_at_block_height | included_in_block_height |
|---|--|
| 10 | 10 |
| 10 | 11 |
| 10 | 12 |

I have looked at your example and realized there is a mistake. It should be like this:

| block_height | height_included |
|---|--|
| 10 | 10 |
| 11 | 10 |
| 11 | 11 |

---

I don't think the suggested naming is better than the existing one.

telezhnaya: > And looks obvious.

I disagree. height_included what? Is height itself included in chunk? (no, I know this because I have domain knowledge. New team member will be doomed here)

telezhnaya: Can you extend your example to 3 blocks or N blocks? Will we have C(N, 2) rows? Why do we need to store it this way?

telezhnaya: included_in_block_height actually solves this concrete issue, it explicitly says that chuck is included in some block height.

khorolets: > included_in_block_height actually solves this concrete issue

OK, I am ready to accept it. What about the `block_height` then?

telezhnaya: Some questions are still unanswered:
Can you extend your example to 3 blocks or N blocks? Will we have C(N, 2) rows? Why do we need to store it this way?

Also, can you explain what `block_height` should be used for? I can't explain this, as a result, obviously, I can't name this.

khorolets: @telezhnaya and I had a call. Recap:

`block_height` identifies which block the given chunk appears in (it might appear in a couple of blocks, though).

We agreed to rename `height_included` into `stored_at_block_height`. This field points to the block height where the chunk can be found on the NEAR Lake S3 bucket (since there is only one copy of the chunk).

Whenever a chunk is missed and `nearcore` includes the previously available, we store this as a duplicate record, but `stored_at_block_height` will always point to the original source.

rpc-server/src/storage.rs
```diff
@@ -29,12 +53,12 @@ impl ScyllaStorageManager for ScyllaDBManager {
...
             get_block_by_hash: Self::prepare_read_query(
                 &scylla_db_session,
-                "SELECT block_height FROM state_indexer.chunks WHERE block_hash = ? LIMIT 1",
+                "SELECT block_height FROM state_indexer.blocks WHERE block_hash = ? LIMIT 1",
             ).await?,
 
             get_block_by_chunk_id: Self::prepare_read_query(
                 &scylla_db_session,
-                "SELECT block_height, shard_id FROM state_indexer.chunks WHERE chunk_hash = ? LIMIT 1",
+                "SELECT stored_at_block_height, shard_id FROM state_indexer.chunks WHERE chunk_hash = ? LIMIT 1",
```
telezhnaya, line 61: Why do we need limit here and above?

telezhnaya (approved):

PR #109, opened 2023-09-01
Improvement (rpc_server): QUERY shadow consistency logging error improvement

kobayurii:
Main idea of this PR to improve QUERY shadow consistency logging error. We have difficulty in finding the right logs for the query method. We have to spend a lot of time looking for the right log. In this PR we have added a suffix with the type of request (`VIEW_CODE`, `VIEW_ACCOUNT` ...) which will help to filter logs by the type of request.


state-indexer/njr.json
khorolets, line 1: Oh, thanks! My bad!

khorolets (approved):

PR #110, opened 2023-09-04
ND-605: query.view_account different results

kobayurii:
1. Since we do queries with the clause WHERE block_height <= X, we need to make sure that the block we are doing a shadow data consistency check for matches the one we got the result for. That's why we are using the block_height from the result.
2. Performance improvement: delete additional query to get block_hash


khorolets (approved):
I totally missed the fact we already have `block_hash` in all these tables 🤦‍♂️

Thank you for improving it!

PR #112, opened 2023-09-06
ND-604: query.function_call different results

kobayurii:
This PR includes improvement for `query:function_call`
1. Added `random_seed` for function_call
2. Moved `max_gas_burnt` into the configuration options
3. Added an error handler to function_call the same as in NEAR RPC so that the error responses coincide.

### DIFF
--- a/rpc-server/src/errors.rs
+++ b/rpc-server/src/errors.rs
@@ -92,3 +92,56 @@ where
         }
     }
 }
+
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum CallFunctionError {
+    #[error("Account ID \"{requested_account_id}\" is invalid")]
+    InvalidAccountId {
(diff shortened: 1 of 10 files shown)


rpc-server/src/errors.rs
```diff
@@ -92,3 +92,56 @@ where
         }
     }
 }
+
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum CallFunctionError {
```
khorolets, line 97: It'd be better to stay consistent and name it `FunctionCallError` (word ordering)

khorolets (commented):
I would appreciate some description even though it is a draft PR, since you've requested my review.

Overall it looks like you've copied logic from `near-vm` or `nearcore`, I don't get the reason yet.

khorolets (approved):

PR #114, opened 2023-09-11
ND-601: Block finality: final issues

kobayurii:


rpc-server/Cargo.toml
```diff
@@ -60,6 +61,8 @@ near-vm-runner = "0.17.0"
 near-vm-logic = "0.17.0"
 near-vm-errors = "0.17.0"
 
+near-lake-framework = { path = "../../near-lake-framework-rs" }
```
khorolets, line 64: Marking it here to make sure we update it with the release of NEAR Lake Framework

rpc-server/src/config.rs
```diff
@@ -1,14 +1,12 @@
...
 
 #[derive(Parser)]
 #[clap(author, version, about, long_about = None)]
 pub struct Opts {
-    // near network rpc url
-    #[clap(long, env = "NEAR_RPC_URL")]
-    pub rpc_url: http::Uri,
+    // Chain ID: testnet or mainnet
+    #[clap(subcommand)]
+    pub chain_id: ChainId,
```
khorolets, line 9: In `LakeConfigBuilder` both `mainnet` and `testnet` are shortcuts to set the proper `s3_bucket_name`. Since we already has it as parameter for ReadRPC, I am sure we don't need `chain_id` in ReadRPC config. Please refactor it accordingly

rpc-server/src/utils.rs
```diff
@@ -28,61 +18,87 @@ where
...
+) -> anyhow::Result<()> {
+    let block = CacheBlock {
+        block_hash: streamer_message.block.header.hash,
+        block_height: streamer_message.block.header.height,
+        block_timestamp: streamer_message.block.header.timestamp,
+        latest_protocol_version: streamer_message.block.header.latest_protocol_version,
+        chunks_included: streamer_message.block.header.chunks_included,
+        state_root: streamer_message.block.header.prev_state_root,
+    };
+    tracing::debug!("Block {:?}", block);
```
khorolets, line 65: This log message is too implicit. And it will be quite noisy. I don't think it makes a lot of sense as we set the metric below which gives us an understanding whether it is working or not. 

```suggestion
```

khorolets (changes requested):
Pretty decent PR, thank you! I left a couple of comments here and there.

rpc-server/Cargo.toml
kobayurii: related from this pr https://github.com/near/near-lake-framework-rs/pull/86

khorolets (approved):

PR #119, opened 2023-09-13
ND-608: query.view_access_key error response mismatch

kobayurii:
In this PR fixed `query.view_access_key` error response mismatch


khorolets (approved):
Great! Thanks!

PR #120, opened 2023-09-13
ND-561: Get rid of all potential data losses

kobayurii:
This PR solves two problems: first, it solves the problem of lost cashe after interruption and second, it handles transaction_hash collision.
 
### **Lost cashe after interruption**
The current logic for collecting data about receipts is the following:

1. We save the data to Scylla only when we have the full picture: tx + all receipts created by this tx.
2. We use a cache to store partially collected receipts
3. If we meet the receipt without previous history in the cache, we ignore it

We use several indexers to fill in the full history, split by time periods → at the beginning of each indexer’s work, we will ignore all receipts for transactions from the previous time period. It could be solved by overlapping the periods, but we never can be sure that overlap is enough.

Moreover, even if we run only one indexer instance from Genesis to the final state, the machine reloads from time to time, and the cache is lost. 

We not only lost the data, we don’t know what was lost.

In this PR we added logic to save transactions in process in the Scylla db to avoid lost data when we restart indexers.

1. Start saving all the data we see in the additional keyspace.
    1. When we meet a new tx with its first receipt_id, we write it to Scylla. We also save it to the local in-memory cache.
    2. When we meet execution_outcome, we also write it to Scylla and to the local in-memory cache.
    
2. Restoring cache after reloading indexer.
    1. Each time the indexer starts, it collects all transactions in the process from Scylla db to the cache. Then, we try to 
    collect data for transactions from the receipts_outcomes table. 
    2. After we restored the cache, we ran normally, collecting the next data as usual.
    
3. Save transactions.
    After successfully saving the transaction into the database we delete all data about it from the `tx_indexer_process`

For transactions in process, we use additional keyspace `tx_indexer_process` with 2 tables:
1. `transactions` 
    (
        transaction_hash varchar,
        block_height varint,
        block_hash varchar,
        transaction_details BLOB,
        PRIMARY KEY (transaction_hash, block_height)
    )
2. `receipts_outcomes`
    (
        transaction_hash varchar,
        block_height varint,
        receipt_id varchar,
        receipt BLOB,
        outcome BLOB,
        PRIMARY KEY (transaction_hash, block_height, receipt_id)
    )

### **Handle transaction_hash collision**
In the current indexer process, we use transaction_hash as keys for transaction collection, which is unacceptable because we have a transaction_hash collision. This can lead to incorrect data collection. To avoid this I introduced a new structure `TransactionKey` which includes `transaction_hash` and `block_height`(the height of the block when we first meet the transaction). This pair guarantees us the uniqueness of transactions.


telezhnaya (commented):
Could you please write the description explaining what are you trying to achieve, how, what do we need to keep in mind while reviewing, etc

telezhnaya (commented):
I think we need to wait for @khorolets to make the final review

tx-indexer/src/config.rs
```diff
@@ -301,6 +341,31 @@ impl ScyllaStorageManager for ScyllaDBManager {
                     VALUES (?, ?)",
             )
             .await?,
+
+            add_transaction_process: Self::prepare_write_query(
```
khorolets, line 345: I would suggest renaming to `process_add_transaction` or even something like `cache_add_transaction` and the similar names below for better readability.

Right now, having the `_process` suffix is misleading.

tx-indexer/src/config.rs
```diff
@@ -264,6 +271,35 @@ impl ScyllaStorageManager for ScyllaDBManager {
...
 
+        scylla_db_session
+            .use_keyspace("tx_indexer_process", false)
+            .await?;
+        scylla_db_session
+            .query(
+                "CREATE TABLE IF NOT EXISTS transactions (
+                transaction_hash varchar,
+                transaction_details BLOB,
+                PRIMARY KEY (transaction_hash)
```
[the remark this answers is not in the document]
khorolets: I double it

tx-indexer/src/config.rs
```diff
@@ -369,4 +434,116 @@ impl ScyllaDBManager {
...
+    ) -> anyhow::Result<()> {
+        let transaction_hash = transaction_details.transaction.hash.clone().to_string();
+        let transaction_details = transaction_details.try_to_vec()?;
+        Self::execute_prepared_query(
+            &self.scylla_session,
+            &self.add_transaction_process,
+            (transaction_hash, transaction_details),
+        )
+        .await?;
+        Ok(())
```
khorolets, line 450: Are you sure we don't want to handle any errors in these functions?

tx-indexer/src/config.rs
```diff
@@ -369,4 +434,116 @@ impl ScyllaDBManager {
...
+                indexer_execution_outcome_with_receipt
+                    .execution_outcome
+                    .try_to_vec()?,
+            ),
+        )
+        .await?;
+        Ok(())
+    }
+
+    pub(crate) async fn get_transactions_in_process(
```
khorolets, line 478: Consider replacing `process` with `progress`. I believe it reflects what we are doing better. @telezhnaya wdyt?

tx-indexer/src/main.rs
```diff
@@ -35,11 +33,17 @@ async fn main() -> anyhow::Result<()> {
...
         .await?,
     );
+    let scylla_session = scylla_db_client.scylla_session().await;
 
     tracing::info!(target: INDEXER, "Generating LakeConfig...");
-    let scylla_session = scylla_db_client.scylla_session().await;
     let config: near_lake_framework::LakeConfig = opts.to_lake_config(&scylla_session).await?;
 
+    tracing::info!(target: INDEXER, "Creating hash storage...");
+    // let tx_collecting_storage = std::sync::Arc::new(storage::hash::HashStorage::new());
```
khorolets, line 42: Don't forget to remove the comment if you don't need this code anymore.

tx-indexer/src/storage/database.rs
```diff
@@ -0,0 +1,321 @@
...
+    transactions_to_save: futures_locks::RwLock<
+        std::collections::HashMap<String, readnode_primitives::CollectingTransactionDetails>,
+    >,
+}
+
+impl HashStorageWithDB {
+    #[allow(unused)]
+    pub(crate) async fn init_storage(
+        scylla_db_client: std::sync::Arc<crate::config::ScyllaDBManager>,
+    ) -> anyhow::Result<Self> {
```
khorolets, line 22: It looks weird to me that `init_storage` function might fail.
Especially, when I look into the reason where it can fail. 
I suggest you to refactor it a bit.
Let's have a function `init_storage` that will initiate the struct instance without a potential failure.

And let's have a "sugar" function like something like `init_with_upload...`. 

BTW looking at the `upload_transaction_with_receipts` first of all I believe we should find a better name to reflect the term "transactions with receipt". And it looks more like a download than an upload. Am I missing something?

khorolets (changes requested):
Overall looks pretty good! Thank you!

I left comments here and there, though. I want to draw your attention to @telezhnaya comment about transaction hash collisions we have to handle.

Also, can you provide some info about the overhead and how these changes affect the BPS speed of the `tx-indexer` in general?

tx-indexer/src/collector.rs
```diff
@@ -1,35 +1,33 @@
...
-    FutureExt, StreamExt,
+    StreamExt,
 };
 use near_indexer_primitives::IndexerTransactionWithOutcome;
 
-use crate::{config, storage};
+use crate::config;
+use crate::storage::base::TxCollectingStorage;
 
 // TODO: Handle known TX hash collision case for mainnet
```
khorolets, line 10: Drop this comment, please while you're at it, it is no longer relevant

khorolets (approved):
Looks pretty legit! I left a comment to drop the irrelevant leftover comment. Please address it before merging 🙏

Issue #117, opened 2023-09-12
`rpc-server` dies from OOM from time to time


khorolets:
It has been a couple of times already when the `rpc-server` has died with out-of-memory code (137). The first time @ecp88 increased the RAM to 32GB (too much), it happened again.

I can confirm that the Docker setup on the VM allows the container to utilize all 32GB of available RAM.

The only place(s) in the `rpc-server` where we might leak memory is the cache:
* __Contract code cache__
* __Contract state cache__
* Latest final block cache
* Genesis cache (constant from the start)

The bolded variants above must be a reason for OOM. 

@kobayurii provided some data about the limitations we have set up there.

* `blocks_cache`
  * size `100_000`
  * data: 
    ```
    pub struct CacheBlock {
        pub block_hash: near_primitives::hash::CryptoHash,
        pub block_height: near_primitives::types::BlockHeight,
        pub block_timestamp: u64,
        pub latest_protocol_version: near_primitives::types::ProtocolVersion,
        pub chunks_included: u64,
    }
    ```
* `contract_code_cache`
  * size `128`
  * data: contract_code from ScyllaDB (bytes)
* `compiled_contract_code_cache`  (using inside `vm_runner` (`vm_runner::precompile_contract`, `vm_runner::run`))
  * size: 128
  * data: `CompiledContract` created in `vm_runner` (bytes)

Alternatively, we could devise a way of clearing that. I want an rpc-server instance to consume 8GB at the top. Ideally, it should be configurable or read from the system, preventing taking more than the server can give.

PR #123, opened 2023-09-29
ND-541: Read RPC server dies with OOM from time to time

kobayurii:
We have a problem that the service crashes from time to time with an OOM error. This is most likely related to caching. I added a memory wrapper around the LRU сaсhe to avoid such situations. I also added the ability to set cache parameters via configuration. To make sense of the situation, I added information to the status method which returns the memory’s information on the server and how much memory cache takes up.

### DIFF
--- a/rpc-server/src/config.rs
+++ b/rpc-server/src/config.rs
@@ -65,28 +65,23 @@ pub struct Opts {
     #[clap(long, env, default_value = "300000000000000")]
     pub max_gas_burnt: near_primitives_core::types::Gas,
 
-    /// Max contract size is limited 4MB
-    /// Configured in the near-core
-    #[clap(long, env, default_value = "4194304")]
-    pub max_contract_size: u64,
-
     /// Max available memory for `block_cache` and `contract_code_cache` in bytes
     /// By default we use all available memory
     /// Example: 1GB = 1073741824 bytes
     #[clap(long, env)]
-    pub limit_memory_cache: Option<u64>,
+    pub limit_memory_cache: Option<usize>,
 
     /// Reserved memory for running the application in bytes
     /// By default we use 256MB (268_435_456 bytes
     #[clap(long, env, default_value = "268435456")]
-    pub reserved_memory: u64,
+    pub reserved_memory: usize,
 
     /// Block cache size in bytes
     /// By default we use 128MB (134_217_728 bytes)
     /// One cache_block size is ≈ 96 bytes
     /// In 128MB we can put 1_398_101 cache_blocks
     #[clap(long, env, default_value = "134217728")]
-    pub block_cache_size: u64,
+    pub block_cache_size: usize,
 }
 
 impl Opts {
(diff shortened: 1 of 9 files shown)


rpc-server/src/config.rs
```diff
@@ -64,6 +64,24 @@ pub struct Opts {
     /// Default value is 300_000_000_000_000
     #[clap(long, env, default_value = "300000000000000")]
     pub max_gas_burnt: near_primitives_core::types::Gas,
+
+    /// Max available memory for `block_cache` and `contract_code_cache` in bytes
```
khorolets, line 68: That's the opposite of good UX, forcing people to provide raw bytes when dealing with gigabytes. Please accept gigabytes here and convert them into bytes later in the code.

Do this for every place where you were going to ask people for bytes.

rpc-server/src/modules/network/methods.rs
```diff
@@ -1,16 +1,55 @@
...
+    let used_memory = sys.used_memory();
+    let blocks_cache = data.blocks_cache.read().unwrap();
+    let contract_code_cache = data.contract_code_cache.read().unwrap();
+    let compiled_contract_code_cache = data
+        .compiled_contract_code_cache
+        .local_cache
+        .read()
+        .unwrap();
+    let status = StatusResponse {
+        total_memory: friendly_memory_size_format(total_memory as usize),
```
khorolets, line 26: You even have something for friendly output :)

rpc-server/src/utils.rs
```diff
@@ -100,6 +101,33 @@ pub async fn update_final_block_height_regularly(
...
+    limit_memory_cache: Option<usize>,
+) -> usize {
+    let sys = System::new_all();
+    let total_memory = sys.total_memory() as usize; // Total memory in bytes
+    let used_memory = sys.used_memory() as usize; // Used memory in bytes
+    let available_memory = total_memory - used_memory - reserved_memory; // Available memory in bytes
+
+    let mem_cache_size = if let Some(limit) = limit_memory_cache {
+        if limit >= available_memory {
+            panic!("Not enough memory to run the server. Available memory: {} bytes, required memory: {} bytes", available_memory, limit);
```
khorolets, line 120: here's another place to use friendly memory size

khorolets (changes requested):
This looks great, except for the inputs in bytes. I left a comment, once addressed we're ready to merge it. Thanks!

kobayurii:
https://github.com/near/read-rpc/issues/117

rpc-server/src/config.rs
```diff
@@ -140,6 +158,6 @@ impl near_primitives::types::CompiledContractCache for CompiledCodeCache {
     }
 
     fn has(&self, key: &near_primitives::hash::CryptoHash) -> std::io::Result<bool> {
-        Ok(self.local_cache.write().unwrap().get(key).is_some())
+        Ok(self.local_cache.write().unwrap().contains(key))
```
khorolets, line 161: Are we ok with that `unwrap` here? Should we handle it, or do we want the app to crash if we cannot unwrap it?

khorolets (approved):
I am approving, though I've noticed some `unwrap` I missed before, decide by yourself what to do with it.

rpc-server/src/config.rs
kobayurii: As far as I know, these errors will be handled in VM Runner.

PR #125, opened 2023-10-16
(tx-indexe) query improvement

kobayurii:
In this PR we rewrite queries to get transactions and get receipts on prepared queries

### DIFF
--- a/tx-indexer/src/config.rs
+++ b/tx-indexer/src/config.rs
@@ -221,6 +221,8 @@ pub(crate) struct ScyllaDBManager {
     add_receipt: PreparedStatement,
     update_meta: PreparedStatement,
 
+    cache_get_transactions: PreparedStatement,
+    cache_get_receipts: PreparedStatement,
     cache_add_transaction: PreparedStatement,
     cache_delete_transaction: PreparedStatement,
     cache_add_receipt: PreparedStatement,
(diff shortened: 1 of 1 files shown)


tx-indexer/src/config.rs
```diff
@@ -503,10 +515,7 @@ impl ScyllaDBManager {
         let mut result = std::collections::HashMap::new();
         let mut rows_stream = self
             .scylla_session
-            .query_iter(
-                "SELECT transaction_details FROM tx_indexer_cache.transactions",
-                &[],
-            )
+            .execute_iter(self.cache_get_transactions.clone(), &[])
```
khorolets, line 518: I think it's better to make these functions on the ScyllaManager side like we did for others to call it ScyllaManager::cache_get_transactions() instead of this mess. Wdyt?

tx-indexer/src/config.rs
```diff
@@ -527,9 +536,12 @@ impl ScyllaDBManager {
         let mut result = vec![];
         let mut rows_stream = self
             .scylla_session
-            .query_iter(
-                "SELECT receipt, outcome FROM tx_indexer_cache.receipts_outcomes WHERE transaction_hash = ? AND block_height = ?",
-                (transaction_key.transaction_hash.clone(), num_bigint::BigInt::from(transaction_key.block_height)),
+            .execute_iter(
```
khorolets, line 539: Same here

khorolets (changes requested):

khorolets (approved):
Disregard my comments, the code is already is the helper functions 🤦‍♂️

PR #127, opened 2023-10-17
add paginator into cache storage

kobayurii:
In this PR we added a paginator in to query to read all transactions in the process. We did it because Scylla db failed when we tried to scan all rows in the table on start.


khorolets (approved):
Can't say I like it. But I won't block it. But we might change it in future

PR #135, opened 2023-11-08
feat: Database agnosticism. Part 2: Add PostgresDB 

kobayurii:
This pull request is a pivotal addition to the project, introducing native support for PostgreSQL as a supported Database Management System (DBMS). This implementation extends the existing functionality to seamlessly integrate and interact with PostgreSQL databases.

**Key Changes:**

1. Addition of PostgreSQL-specific database interaction modules.
2. Implementation of necessary queries and data handling methods tailored for PostgreSQL compatibility.
3. Configuration enhancements to accommodate PostgreSQL connection parameters.

**Importance of this Update:**

1. **_Expanded Database Compatibility:_** With the inclusion of PostgreSQL support, this update significantly widens the scope of database options available for users, catering to those preferring PostgreSQL as their DBMS.
2. **_Enhanced Flexibility:_** Developers now have the ability to seamlessly utilize the project with either ScyllaDB or PostgreSQL, ensuring increased flexibility and adaptability to diverse project requirements.
3. **_Increased User Base:_** By supporting PostgreSQL, this pull request makes the project more accessible to a broader audience, potentially attracting users who rely on PostgreSQL in their projects.

By incorporating native support for PostgreSQL, this pull request not only enriches the project's capabilities but also fortifies its position as a versatile solution, empowering users with greater freedom in selecting their preferred database system.

---

Additional context by @khorolets:

Heads up! ReadRPC is not expected to handle all the load using PostgreSQL instead of ScyllaDB if it watches all the data related to all the accounts. 

The main purpose of this PR is to bring the support of PostgreSQL for the next step to make a data filtering, that will allow node operator to configure the setup to track the data related to the limited set of accounts (this PR does not include this feature).


database/diesel.toml
```diff
@@ -0,0 +1,9 @@
+# For documentation on how to configure this file,
+# see https://diesel.rs/guides/configuring-diesel-cli
+
+[print_schema]
+file = "src/postgres/schema.rs"
+custom_type_derives = ["diesel::query_builder::QueryId"]
```
khorolets, line 6: Could you clarify what does it mean?

database/src/base/mod.rs
```diff
@@ -5,15 +5,49 @@ pub use crate::base::state_indexer::StateIndexerDbManager;
...
+/// `keepalive_interval: Option<u64>`
+/// ScyllaDB keepalive interval
+///
+/// `max_retry: u8`
+/// Max retry count for ScyllaDB if `strict_mode` is `false`
+///
+/// `strict_mode: bool`
+/// Attempts to store data in the database should be infinite to ensure no data is missing.
+/// Disable it to perform a limited write attempts (`max_retry`)
+/// before skipping giving up and moving to the next piece of data
```
khorolets, line 27:
```suggestion
/// before giving up, and moving to the next piece of data
```

database/src/base/mod.rs
```diff
@@ -5,15 +5,49 @@ pub use crate::base::state_indexer::StateIndexerDbManager;
...
+    #[cfg(feature = "scylla_db")]
+    pub preferred_dc: Option<String>,
+    #[cfg(feature = "scylla_db")]
+    pub keepalive_interval: Option<u64>,
+    #[cfg(feature = "scylla_db")]
+    pub max_retry: u8,
+    #[cfg(feature = "scylla_db")]
+    pub strict_mode: bool,
+    #[cfg(feature = "postgres_db")]
+    pub database_name: Option<String>,
```
khorolets, line 42: BTW why is `database_name` optional?

database/src/postgres/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[async_trait::async_trait]
+pub trait PostgresStorageManager {
+    async fn create_pool(
+        database_url: &str,
+        database_user: Option<&str>,
+        database_password: Option<&str>,
+        database_options: crate::AdditionalDatabaseOptions,
+    ) -> anyhow::Result<PgAsyncPool> {
+        let connecting_url = if database_url.starts_with("postgres://") {
```
khorolets, line 20: `connection_string` would be a better name. I don't think the continuous tense is a good choice for variables like this.

database/src/postgres/models.rs
```diff
@@ -0,0 +1,569 @@
...
+pub struct StateChangesData {
+    pub account_id: String,
+    pub block_height: bigdecimal::BigDecimal,
+    pub block_hash: String,
+    pub data_key: String,
+    pub data_value: Option<Vec<u8>>,
+}
+
+impl StateChangesData {
+    pub async fn save(&self, mut conn: crate::postgres::PgAsyncConn) -> anyhow::Result<()> {
```
khorolets, line 17: The verb "save" has two meanings in this context:
- Save as new in DB (insert)
- Save after changes (update)

Please, consider renaming to something straightforward for our own sake.

database/src/postgres/models.rs
```diff
@@ -0,0 +1,569 @@
...
+        Ok(())
+    }
+
+    pub async fn get_state_key_value(
+        mut conn: crate::postgres::PgAsyncConn,
+        account_id: &str,
+        block_height: u64,
+        data_key: String,
+    ) -> anyhow::Result<Option<Vec<u8>>> {
+        let resp = state_changes_data::table
```
khorolets, line 32: use the full word "response", please

database/src/postgres/models.rs
```diff
@@ -0,0 +1,569 @@
...
+    pub async fn save(&self, mut conn: crate::postgres::PgAsyncConn) -> anyhow::Result<()> {
+        diesel::insert_into(state_changes_data::table)
+            .values(self)
+            .on_conflict_do_nothing()
+            .execute(&mut conn)
+            .await?;
+        Ok(())
+    }
+
+    pub async fn get_state_key_value(
```
khorolets, line 26: I would expect to use it like this:

```rust
StateChangesData::get(...)
```

For me, `StateChangesData::get_state_key_value(...)` looks  
repetitive. Since, you're not just returning the model instance in this particular case I'd suggest something like `get_data_value`

For other models returning `Self`, I'd use `get()` as a name for the getter. What do you think?

khorolets (changes requested):
Overall looks great! I've requested a few changes/improvements and we're good to go.

database/src/postgres/models.rs
kobayurii: Diesel has already implemented a method called `get` which is used to get the values of a particular field. I also think that the current names reflect the essence of what is returned in queries.

khorolets: I definitely can live with that. I hope we do not end up in confusion later. I'll leave it to you. Decide what to do with this suggestion.

database/src/base/mod.rs
kobayurii: we can use 2 types of configuration: 
1. specify all parameters individually 
```
DATABASE_URL="localhost"
DATABASE_USER="postgres"
DATABASE_PASSWORD="password"
DATABASE_NAME="near_data 
```
2. specify the connection string at once 
```
DATABASE_URL=postgres://postgres:password@localhost/near_data 
```

database/diesel.toml
kobayurii: It is generated automatically by Diesel CLI.
https://diesel.rs/guides/configuring-diesel-cli

database/src/base/mod.rs
khorolets: cool, missed that 👍

khorolets (approved):

Issue #115, opened 2023-09-12
`query.function_call` ReadRPC returns error while NEAR RPC succeeds


khorolets:
### The request

```
{
  "jsonrpc": "2.0",
  "id": "dontcare",
  "method": "query",
  "params": {
    "request_type": "call_function",
    "block_id": 100523214,
    "account_id": "ofpcore.near",
    "method_name": "get_stages_max_number",
    "args_base64": "eyJwcm9qZWN0X2lkIjogIjE2NjE5NDk3NjkxOTIifQ=="
  }
}
```

### ReadRPC response
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32000,
        "message": "Server error",
        "data": "Function call returned an error: wasm execution failed with error: CompilationError(PrepareError(Deserialization))",
        "name": "HANDLER_ERROR",
        "cause": {
            "info": {
                "block_hash": "74ENVfd6tHDCoKEfipcFWMDFyeJQR1EMX1J86oF2smtZ",
                "block_height": 100523214,
                "vm_error": "wasm execution failed with error: CompilationError(PrepareError(Deserialization))"
            },
            "name": "CONTRACT_EXECUTION_ERROR"
        }
    },
    "id": "dontcare"
}
```

### NEAR RPC response
```
{
    "jsonrpc": "2.0",
    "result": {
        "block_hash": "74ENVfd6tHDCoKEfipcFWMDFyeJQR1EMX1J86oF2smtZ",
        "block_height": 100523214,
        "logs": [],
        "result": [
            34,
            51,
            50,
            34
        ]
    },
    "id": "dontcare"
}
```

We might need help from the contract runtime team. Perhaps there are some corner cases in the way we run the contract that lead to 

```
"wasm execution failed with error: CompilationError(PrepareError(Deserialization))"
```

miraclx:
Let's update the rust base images in docker to `>0.70` and see what happens then. I suspect the contract deployed on `ofpcore.near` might've been built with a version of rust that includes instructions from the signed extensions proposal.

Context: https://github.com/near/nearcore/pull/9140

kobayurii:
> Let's update the rust base images in docker to `>0.70` and see what happens then. I suspect the contract deployed on `ofpcore.near` might've been built with a version of rust that includes instructions from the signed extensions proposal.
> 
> Context: [near/nearcore#9140](https://github.com/near/nearcore/pull/9140)

[...]

PR #139, opened 2023-11-27
query.function_call ReadRPC returns error while NEAR RPC succeeds

kobayurii:
This pull request encompasses several crucial updates and resolves specific issues.

1. update dependencies
2. up rust version
3. query.function_call ReadRPC returns error while NEAR RPC succeeds

Fixed an issue with the query.function_call ReadRPC call that used to inconsistently return errors while NEAR RPC was successful. This was because a version of nearcore less than 1.35 did not support running contracts built on rust 1.70+.


kobayurii:
https://github.com/near/read-rpc/issues/115

Cargo.toml
```diff
@@ -8,3 +10,14 @@ members = [
...
+
+[patch.crates-io]
+near-chain-configs = { git = 'https://github.com/near/nearcore.git', branch = '1.36.0' }
+near-crypto = { git = 'https://github.com/near/nearcore.git', branch = '1.36.0' }
+near-jsonrpc-primitives = { git = 'https://github.com/near/nearcore.git', branch = '1.36.0' }
+near-indexer-primitives = { git = 'https://github.com/near/nearcore.git', branch = '1.36.0' }
+near-primitives = { git = 'https://github.com/near/nearcore.git', branch = '1.36.0' }
+near-vm-runner = { git = 'https://github.com/near/nearcore.git', branch = '1.36.0' }
+
+near-jsonrpc-client = { git = 'https://github.com/kobayurii/near-jsonrpc-client-rs.git', branch = '0.7.0' }
```
khorolets, line 22: Add comments around these two dependencies explaining why we use your forks instead of going through the PRs process in those projects, please.

khorolets (approved):
LGTM! I've asked to leave a comment around the dependency so we handle the updates when it is possible.

Issue #144, opened 2023-12-12
`view_state` should not be limited to 101 keys


frol:
With the current implementation:

https://github.com/near/read-rpc/blob/c1e12b8a690e6d87b0c1f4a16c3863f06744b758/rpc-server/src/modules/queries/utils.rs#L60-L62

`view_state` method truncates the output to 101 keys. How heavy would it be to allow up to 10k keys and, ideally, implement pagination?

cc @khorolets

frol:
Resolved with #145.

I have just tested it with:

```
near contract view-storage poolv1.near all as-json network-config mainnet now
```

PR #145, opened 2023-12-13
remove `view_state` limit

kobayurii:
**Changes Implemented**

**1. Removed Key Limit:** The previously imposed limitation on the number of keys in a state has been eliminated. This means our system can now handle a larger volume of keys without restrictions.
**2. Optimization of Queries:** We've implemented optimizations for both key and value queries. This enhancement ensures improved efficiency and performance when retrieving information related to keys and their corresponding values.

**Temporary Solution**
It's important to note that while these changes represent a substantial improvement, this solution is temporary. Although the system now allows for handling a greater number of keys, there's no absolute guarantee for handling exceptionally large datasets, such as those in the gigabyte range.

**Future Steps**

**1. Pagination Implementation:** The upcoming step will introduce new endpoints designed explicitly for pagination. These endpoints will facilitate the handling of large states by allowing the data to be retrieved in manageable chunks or pages.
**2. Guaranteed Retrieval for Large States**: These new endpoints will guarantee the retrieval of all keys, even for states with massive data volumes, ensuring that clients can reliably access and process extensive datasets without constraints.

This pull request lays the groundwork for future improvements, acknowledging the current enhancements while also outlining a clear path for addressing the handling of exceptionally large datasets in the upcoming iterations of the system.

### DIFF
--- a/database/src/base/rpc_server.rs
+++ b/database/src/base/rpc_server.rs
@@ -31,7 +31,10 @@ pub trait ReaderDbManager {
         account_id: &near_primitives::types::AccountId,
         block_height: near_primitives::types::BlockHeight,
         key_data: readnode_primitives::StateKey,
-    ) -> anyhow::Result<readnode_primitives::StateValue>;
+    ) -> (
+        readnode_primitives::StateKey,
+        readnode_primitives::StateValue,
+    );
 
     /// Returns the near_primitives::account::Account at the given block height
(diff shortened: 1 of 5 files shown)


kobayurii:
https://github.com/near/read-rpc/issues/144

khorolets (approved):
LGTM!

Issue #159, opened 2024-01-17
(state-indexer): Bug with indexing epoch validators info


kobayurii:
Information about validators of the epoch may change in the course of Near operation. 
StakeAction(receipt action) serves as a trigger for changes. 
The indexer ignores this and indexes validator information at the very beginning of the epoch. 
This leads to the fact that the server has incomplete information about validators of the epoch. 
It is necessary to improve the indexing process:
1. Configure full saving of information about validators at the end of the epoch.
2. Configure the update of information about validators by Stake action.
3. Add information about the last block of the epoch

PR #160, opened 2024-01-17
(state-indexer): Bug with indexing epoch validators info

kobayurii:
The current implementation faces a challenge where information about validators during an epoch in Near operations may undergo changes. The StakeAction, acting as a trigger for these changes, is not properly considered by the indexer. As a result, the indexer neglects to update validator information when StakeAction occurs, leading to incomplete and inaccurate data about validators for that epoch.

In this PR:
1. Configured the system to save comprehensive information about validators at the conclusion of each epoch. This ensures that the indexer captures and indexes the most up-to-date data about validators, eliminating the issue of incomplete information.
2. Added functionality to include information about the last block of the epoch in the indexing process. This addition ensures that the server maintains a complete record of validator information, providing a comprehensive view of the epoch.

### DIFF
--- a/database/src/postgres/rpc_server.rs
+++ b/database/src/postgres/rpc_server.rs
@@ -1,8 +1,10 @@
-use crate::postgres::PostgresStorageManager;
-use crate::AdditionalDatabaseOptions;
+use std::str::FromStr;
+
 use bigdecimal::ToPrimitive;
 use borsh::{BorshDeserialize, BorshSerialize};
-use std::str::FromStr;
+
+use crate::postgres::PostgresStorageManager;
+use crate::AdditionalDatabaseOptions;
 
 pub struct PostgresDBManager {
     pg_pool: crate::postgres::PgAsyncPool,
--- a/database/src/postgres/rpc_server.rs
+++ b/database/src/postgres/rpc_server.rs
@@ -2,6 +2,7 @@ use crate::postgres::PostgresStorageManager;
 use crate::AdditionalDatabaseOptions;
 use bigdecimal::ToPrimitive;
 use borsh::{BorshDeserialize, BorshSerialize};
+use std::str::FromStr;
 
 pub struct PostgresDBManager {
     pg_pool: crate::postgres::PgAsyncPool,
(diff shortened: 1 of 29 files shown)


kobayurii:
https://github.com/near/read-rpc/issues/159

database/src/postgres/rpc_server.rs
```diff
@@ -2,6 +2,7 @@ use crate::postgres::PostgresStorageManager;
 use crate::AdditionalDatabaseOptions;
 use bigdecimal::ToPrimitive;
 use borsh::{BorshDeserialize, BorshSerialize};
+use std::str::FromStr;
```
khorolets, line 5: Please improve the import piece according to the style guides:

1. std goes first
2. External goes next
3. Current crates 
4. Mods

rpc-server/src/modules/blocks/utils.rs
```diff
@@ -97,8 +97,11 @@ pub async fn fetch_block_from_cache_or_get(
         near_primitives::types::BlockReference::Finality(_) => {
             // Returns the final_block_height for all the finalities.
             Ok(data
-                .final_block_height
-                .load(std::sync::atomic::Ordering::SeqCst))
+                .final_block_info
+                .read()
+                .unwrap()
```
khorolets, line 102: I don't like this unhandled, unrecoverable `unwrap`. Can you handle it?

rpc-server/src/modules/network/methods.rs
```diff
@@ -48,8 +47,11 @@ pub async fn status(
         ),
 
         final_block_height: data
-            .final_block_height
-            .load(std::sync::atomic::Ordering::SeqCst),
+            .final_block_info
+            .read()
+            .unwrap()
```
khorolets, line 52: And this one

rpc-server/src/utils.rs
```diff
@@ -129,16 +142,31 @@ async fn handle_streamer_message(
         state_root: streamer_message.block.header.prev_state_root,
         epoch_id: streamer_message.block.header.epoch_id,
     };
-    final_block_height.store(block.block_height, std::sync::atomic::Ordering::SeqCst);
+
+    // let mut finale_block_info_lock = finale_block_info.write().await;
```
khorolets, line 146: Do you need these lines to be comments?

rpc-server/src/modules/blocks/mod.rs
```diff
@@ -12,3 +12,35 @@ pub struct CacheBlock {
     pub state_root: near_primitives::hash::CryptoHash,
     pub epoch_id: near_primitives::hash::CryptoHash,
 }
+
+#[derive(Debug)]
+pub struct FinaleBlockInfo {
```
khorolets, line 17: It should be Final without an `e`

> A "finale" is always a noun, and refers to the end of a performance of some kind

khorolets (changes requested):

database/src/postgres/rpc_server.rs
kobayurii: Unfortunately I can't change the order of the imports. This is handled by `cargo fmt`.

khorolets: You need to leave an empty line between the groups in that case `fmt` won't reorder it

kobayurii: Thank you!

khorolets (approved):

PR #170, opened 2024-02-09
Add documentation

kobayurii:
In this pr, we extend documentation and add a description about rpc methods state and add documentation about custom rpc methods.


khorolets (approved):
Great! Thanks!

PR #174, opened 2024-02-21
Get transaction and receipts from tx-indexer process cache

kobayurii:
1. Added mechanism to get transaction and receipts from tx-indexer process cache
2. Remove cache memory storage from tx-indexer


khorolets (approved):
LGTM! Great work!

Issue #176, opened 2024-02-22
Additional metrics to measure proxy calls (archival)


khorolets:
We got an internal request to measure the number of proxy calls we make.

The main goal is to identify how hard we're touching the ArchivalRPC to judge whether all the proxies we do in this regard are reasonable. Thus, we could adjust the logic (if necessary) to reduce the number of such calls.

kobayurii:
Proxying to archive nodes we have only 2 cases:
1. When we use `shadow_data_consistency`.
2. In the QUERY.VIEW_STATE method with the `include_proof` flag. 

In the first case, we don't need to worry about calling archive nodes because the feature `shadow_data_consistency` is used only for debugging the service and **should not be enabled on the production server**.

In the other case, I added an `archive_proxy_query_view_state_with_include_proofs` metric which counts the number of requests redirected to archive nodes.

kobayurii:
https://github.com/near/read-rpc/pull/177

PR #177, opened 2024-02-22
Additional metrics to measure proxy calls (archival)

kobayurii:
1. Rewrite the method to get genesis_block from `lake_framework` to avoid the call to archive nodes.
2. Added `archive_proxy_query_view_state_with_include_proofs` metric to calculate count of requests to the archive nodes
3. Added separation of calls in the `QUERY.VIEW_STATE` method with the `include_proof` flag into regular and archive nodes


kobayurii:
https://github.com/near/read-rpc/issues/176

khorolets (approved):
Great!

Issue #188, opened 2024-02-28
Enhance `state-indexer` compatibility with NEAR Lake and NEAR Indexer Framework


khorolets:
To improve the efficiency and responsiveness of ReadRPC, we aim to make the `state-indexer` compatible with both NEAR Lake and NEAR Indexer Framework. NEAR Lake-based indexers are known for their lightweight design, cost-effectiveness, and ease of maintenance. However, they inherently introduce a ~3-second delay in data transfer to clients via S3. To address this, we propose developing different state-indexer binaries: one for NEAR Lake and another for the NEAR Indexer Framework. This approach will allow us to maintain the advantages of NEAR Lake while significantly reducing data transfer delays by leveraging a `nearcore`-based `state-indexer`. 

The ultimate goal is to provide ReadRPC and regular RPC node users with the most recent data, mimicking the immediacy expected from NEAR's native RPC.

khorolets:
I've realized that having the `state-indexer` built on top of `nearcore` in the ReadRPC fleet of machines would allow us to redirect write-requests or regular RPC proxies to that node, offloading the Pagoda's RPC service a bit 🤔

kobayurii:
https://github.com/near/read-rpc/pull/189

kobayurii:
https://github.com/near/read-rpc/pull/195

PR #189, opened 2024-02-29
Enhance state-indexer compatibility with NEAR Lake and NEAR Indexer Framework

kobayurii:


kobayurii:
https://github.com/near/read-rpc/issues/188

kobayurii:
https://github.com/near/read-rpc/pull/196

Issue #186, opened 2024-02-28
🔷 [ProjectTracking] Implement support for `optimistic` block finality


khorolets:
# Goals

## Background

Recently we've added additional metrics to measure the number of proxy requests from the ReadRPC to native one. Additionally, we have a metric that counts how many of the requests are for the blocks with optimistic finality. This finality is not yet supported in the ReadRPC. The metrics revealed that a significant portion of the requests are for the optimistic blocks - 85% of all requests. This is a significant portion of the requests that could be served by the ReadRPC if it supported optimistic blocks. Without the optimistic finality support, the ReadRPC is not showing it's potential as a regular RPC for the NEAR blockchain.

I haven't found any documentation about optimistic finality except for this short mention in the NEAR RPC documentation:
* https://docs.near.org/api/rpc/setup#using-finality-param

## Why should NEAR One work on this

Enabling optimistic block support in ReadRPC presents a significant opportunity to enhance dApp responsiveness and reduce the load on NEAR's native RPC. This is crucial for applications that require the most up-to-date blockchain data to ensure dynamic and responsive user interactions.

## What needs to be accomplished

* Modify the NEAR Indexer Framework to support streaming of optimistic blocks.
* Introduce a caching solution for storing data from optimistic blocks in the ReadRPC project.
* Ensure the existing finalized blocks flow remains unaffected while integrating optimistic block processing.

## Main use case

Improving data responsiveness for dApps by providing quicker access to blockchain data through optimistic blocks, thereby enhancing user experience and reducing the load on native RPC services.

## Links to external documentation and discussions

* https://docs.near.org/api/rpc/setup#using-finality-param

## Estimated effort
Engineers assigned to this initiative: @khorolets and @kobayurii

The initial effort estimate is about 1-1.5 PM (person months)

## Assumptions
This project does not make specific assumptions.

## Pre-requisites
N/A

## Out of scope
N/A

## Task list:
```[tasklist]
### Tasks
- [ ] https://github.com/near/read-rpc/issues/188
- [ ] https://github.com/near/read-rpc/issues/199
- [ ] https://github.com/near/read-rpc/issues/200
```

khorolets:
UPDATE 2022-03-11

For the last week, we have made good progress. The PR that introduces support for the optimistic block finalities is under review. Ref: https://github.com/near/read-rpc/pull/197

We aim to wrap up the PR earlier this week and release version 0.2.0 of the ReadRPC for deployment on the testnet as soon as possible.

[...]

PR #191, opened 2024-03-01
Implement support for optimistic block finality

kobayurii:


kobayurii:
https://github.com/near/read-rpc/issues/186

kobayurii:
https://github.com/near/read-rpc/pull/197

PR #196, opened 2024-03-06
add near-state-indexer

kobayurii:


kobayurii:
https://github.com/near/read-rpc/issues/188

khorolets (approved):

PR #197, opened 2024-03-07
Implement support for optimistic block finality

kobayurii:
In this pull request, we introduce support optimistic blocks. 

Recently we've added additional metrics to measure the number of proxy requests from the ReadRPC to the native one. Additionally, we have a metric that counts how many of the requests are for the blocks with optimistic finality. This finality is not yet supported in the ReadRPC. The metrics revealed that a significant portion of the requests are for the optimistic blocks - 85% of all requests.

Implementation of `optimistic_blocks` support will help to reduce the number of proxied requests and help to optimize and speed up the service.

### DIFF
--- a/rpc-server/src/modules/blocks/mod.rs
+++ b/rpc-server/src/modules/blocks/mod.rs
@@ -278,13 +278,13 @@ impl BlockInfo {
 }
 
 #[derive(Debug)]
-pub struct FinalityBlocksInfo {
+pub struct BlocksInfoByFinality {
     pub final_block: BlockInfo,
     pub optimistic_block: BlockInfo,
     pub current_validators: near_primitives::views::EpochValidatorInfo,
 }
 
-impl FinalityBlocksInfo {
+impl BlocksInfoByFinality {
     pub async fn new(
         near_rpc_client: &crate::utils::JsonRpcClient,
         blocks_cache: &std::sync::Arc<
--- a/rpc-server/src/modules/blocks/mod.rs
+++ b/rpc-server/src/modules/blocks/mod.rs
@@ -37,16 +37,19 @@ pub struct BlockInfo {
 }
 
 impl BlockInfo {
+    // Create new BlockInfo from BlockView. this method is useful only for start rpc-server.
     pub async fn new_from_block_view(block_view: near_primitives::views::BlockView) -> Self {
         Self {
             block_cache: CacheBlock::from(&block_view),
             stream_message: near_indexer_primitives::StreamerMessage {
                 block: block_view,
-                shards: vec![],
+                shards: vec![], // We left shards empty because block_view doesn't contain shards.
             },
         }
     }
 
+    // Create new BlockInfo from StreamerMessage.
+    // This is using to update final and optimistic blocks regularly.
     pub async fn new_from_streamer_message(
         stream_message: near_indexer_primitives::StreamerMessage,
     ) -> Self {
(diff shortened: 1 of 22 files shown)


kobayurii:
https://github.com/near/read-rpc/issues/186

rpc-server/src/modules/blocks/mod.rs
```diff
@@ -28,9 +30,237 @@ impl From<near_primitives::views::BlockView> for CacheBlock {
...
+    pub stream_message: near_indexer_primitives::StreamerMessage,
+}
+
+impl BlockInfo {
+    pub async fn new_from_block_view(block_view: near_primitives::views::BlockView) -> Self {
+        Self {
+            block_cache: CacheBlock::from(&block_view),
+            stream_message: near_indexer_primitives::StreamerMessage {
+                block: block_view,
+                shards: vec![],
```
khorolets, line 45: Please leave a comment explaining why it is empty here.

rpc-server/src/modules/blocks/mod.rs
```diff
@@ -28,9 +30,237 @@ impl From<near_primitives::views::BlockView> for CacheBlock {
...
+
+    pub async fn block_view(&self) -> near_primitives::views::BlockView {
+        self.stream_message.block.clone()
+    }
+
+    pub async fn shards(&self) -> Vec<near_indexer_primitives::IndexerShard> {
+        self.stream_message.shards.clone()
+    }
+
+    pub async fn account_changes_in_block(
```
khorolets, line 67: Please leave doc-strings explaining the reasoning behind these methods here.

rpc-server/src/modules/queries/methods.rs
```diff
@@ -241,26 +241,50 @@ async fn view_account(
         account_id,
         block.block_height
     );
-
-    let account = data
-        .db_manager
-        .get_account(account_id, block.block_height)
+    if let Ok(result) = data
```
khorolets, line 244: Put a comment explaining the logic of relying on the Err for non-optimistic queries for future us.

khorolets: Or break down so different functions as discussed on the call

rpc-server/src/modules/queries/methods.rs
```diff
@@ -372,16 +451,29 @@ async fn view_state(
...
-        fetch_state_from_db(&data.db_manager, account_id, block.block_height, prefix)
-            .await
-            .map_err(|_err| {
-                near_jsonrpc_primitives::types::query::RpcQueryError::UnknownAccount {
-                    requested_account_id: account_id.clone(),
-                    block_height: block.block_height,
-                    block_hash: block.block_hash,
-                }
-            })?;
+    let optimistic_data = data
```
khorolets, line 454: If I requested Final, this piece anyway will return me optimistic.

rpc-server/src/modules/queries/utils.rs
```diff
@@ -115,16 +119,36 @@ pub async fn fetch_state_from_db(
         prefix,
     );
     let state_from_db = get_state_keys_from_db(db_manager, account_id, block_height, prefix).await;
-    if state_from_db.is_empty() {
+    if state_from_db.is_empty() && optimistic_data.is_empty() {
```
khorolets, line 122: I wouldn't say I like the idea of putting this logic here. I agree to have it here for now but consider adding a TODO to refactor it later.

khorolets (changes requested):
👍 Looks great! We agreed about some fixes and corrections to be done

khorolets (approved):
Looks good enough for me! Let's start planning the roll-up 👍

PR #207, opened 2024-03-18
add RUST_LOG=info as a default directive

kobayurii:
In this PR added RUST_LOG=info as a default directive and added more info logs for the near-state-indexer on the start.


khorolets (approved):

kobayurii:
https://github.com/near/read-rpc/pull/208

PR #216, opened 2024-03-20
fix: Fix Dockerfile for read-rpc server

khorolets:
To fix the problem with parsing, we had to depend on additional parts of `nearcore`, which require redundant dependency. Right now we need to make the ReadRPC working


kobayurii (approved):

kobayurii:
Nice! Thx!

PR #222, opened 2024-03-26
add warning when we could't borsh dezerialize tx deteils

kobayurii:
For some unknown reason, tx-indexer saves invalid data for transactions. We want to avoid these problems and get more information. That's why we added a mechanism to check and resave transactions. We have also expanded logging so that we can determine the causes in more detail.

### DIFF
--- a/database/src/base/tx_indexer.rs
+++ b/database/src/base/tx_indexer.rs
@@ -8,6 +8,10 @@ pub trait TxIndexerDbManager {
         signer_id: &str,
     ) -> anyhow::Result<()>;
 
+    // This function is used to validate that the transaction is saved correctly.
+    // For some unknown reason, tx-indexer saves invalid data for transactions. 
+    // We want to avoid these problems and get more information. 
+    // That's why we added this method.
     async fn validate_saved_transaction_deserializable(
         &self,
         transaction_hash: &str,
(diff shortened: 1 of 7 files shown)


khorolets (approved):

database/src/base/tx_indexer.rs
```diff
@@ -6,6 +6,12 @@ pub trait TxIndexerDbManager {
         block_height: u64,
     ) -> anyhow::Result<()>;
 
+    async fn check_transaction_save_correctly(
```
khorolets, line 9: My suggestion to rename this function into `validate_saved_transaction_deserializable` to be more explicit of what exactly we want to check. Add comments for future us about the problem we solve by having this.

tx-indexer/src/collector.rs
```diff
@@ -292,11 +293,57 @@ async fn save_transaction_details(
...
+                .await
+            {
+                tracing::warn!(
+                    target: crate::INDEXER,
+                    "Failed to validate transaction {} \n{:#?}",
+                    transaction_hash,
+                    err
+                );
+                if attempts >= 3 {
+                    return false;
```
khorolets, line 321: Please add a warning here with the transaction hash so we can see in logs that we attempted 3 times and failed

khorolets (changes requested):

khorolets (approved):
LGTM! Let's hunt that issue down

PR #225, opened 2024-04-08
Update nearcore to 1.39.0

kobayurii:
This PR includes changes:

1. Update nearcore dependencies to 1.39.0 
2. Improvement logic to set and get finalities blocks from Redis

### DIFF
--- a/near-state-indexer/src/main.rs
+++ b/near-state-indexer/src/main.rs
@@ -1,5 +1,6 @@
 use clap::Parser;
 use futures::StreamExt;
+use near_indexer::near_primitives;
 
 use crate::configs::Opts;
 use near_indexer::near_primitives::hash::CryptoHash;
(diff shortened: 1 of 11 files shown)


near-state-indexer/src/main.rs
```diff
@@ -70,14 +70,14 @@ async fn handle_streamer_message(
         &indexer_config,
     );
 
-    let published_streamer_message_feature =
-        utils::publish_streamer_message("final_block", &streamer_message, redis_client);
+    let update_block_streamer_message_feature =
```
khorolets, line 73: I believe you meant fUture
```suggestion
    let update_block_streamer_message_future =
```

near-state-indexer/src/utils.rs
```diff
@@ -59,16 +59,16 @@ pub(crate) async fn fetch_status(
         .await??)
 }
 
-pub(crate) async fn publish_streamer_message(
-    topic: &str,
+pub(crate) async fn update_block_streamer_message(
+    block_type: &str,
```
khorolets, line 63: I really don't like the approach of using error-prone `&str` in the world of Rust-lang where we have fancy `enum`s that can prevent us from

rpc-server/src/utils.rs
```diff
@@ -233,81 +244,83 @@ pub async fn check_updating_optimistic_block_regularly(
...
+                )
+                .await
+                {
+                    tracing::error!("Error to handle_streamer_message: {:?}", err);
                 }
             }
-        } else {
-            tracing::info!("Redis `final_block` message: {:?}", message);
+            Err(err) => {
+                tracing::error!("Error to get finale block from redis: {:?}", err);
```
khorolets, line 284:
```suggestion
                tracing::error!("Error to get final block from redis: {:?}", err);
```

khorolets (changes requested):

Issue #228, opened 2024-04-11
(rpc-server): `UNKNOWN_BLOCK` error response mismatch


khorolets:
The request (HTTPie)

```
http post https://archival-rpc.mainnet.near.org/ jsonrpc=2.0 id=qwe method=block params:='{"block_id": 57730442}'
```

### Expected response (native NEAR RPC)

```
{
    "error": {
        "cause": {
            "info": {},
            "name": "UNKNOWN_BLOCK"
        },
        "code": -32000,
        "data": "DB Not Found Error: BLOCK HEIGHT: 57730442 \n Cause: Unknown",
        "message": "Server error",
        "name": "HANDLER_ERROR"
    },
    "id": "qwe",
    "jsonrpc": "2.0"
}
```

### Actual response (mismatch)

```
{
    "error": {
        "cause": {
            "info": {},
            "name": "UNKNOWN_BLOCK"
        },
        "code": -32000,
[...]

PR #229, opened 2024-04-11
(rpc-server): UNKNOWN_BLOCK error response mismatch

kobayurii:
This PR includes improvements to the error text when receiving a block. 
This is to maintain full compatibility with the native NEAR RPC.

issue: https://github.com/near/read-rpc/issues/228


khorolets (approved):
Awesome! Thanks!

PR #232, opened 2024-04-18
final block handler improvement

kobayurii:
We need to update final block from Redis and Lake because we can't be sure that Redis has the latest block.
Lake can be used as a backup source

### DIFF
--- a/rpc-server/src/utils.rs
+++ b/rpc-server/src/utils.rs
@@ -279,7 +279,7 @@ pub async fn update_optimistic_block_regularly(
             }
         }
 
-        // When optimistic block is not updated or optimistic block less than final block
+        // When an optimistic block is not updated, or it is lower than the final block
         // we need to mark that optimistic updating is not working
         if crate::metrics::OPTIMISTIC_BLOCK_HEIGHT.get() <= crate::metrics::FINAL_BLOCK_HEIGHT.get()
             && !crate::metrics::OPTIMISTIC_UPDATING.is_not_working()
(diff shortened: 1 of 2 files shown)


rpc-server/src/utils.rs
```diff
@@ -319,18 +261,43 @@ pub async fn update_optimistic_block_regularly(
...
+                            .expect("Invalid optimistic block height"),
+                    );
+                }
             }
             Err(err) => {
                 tracing::error!("Error to get optimistic block from redis: {:?}", err);
             }
         }
+
+        // When optimistic block is not updated or optimistic block less than final block
```
khorolets, line 282:
```suggestion
        // When an optimistic block is not updated, or it is lower than the final block
```

rpc-server/src/utils.rs
```diff
@@ -319,18 +261,43 @@ pub async fn update_optimistic_block_regularly(
...
+        if crate::metrics::OPTIMISTIC_BLOCK_HEIGHT.get() <= crate::metrics::FINAL_BLOCK_HEIGHT.get()
+            && !crate::metrics::OPTIMISTIC_UPDATING.is_not_working()
+        {
+            tracing::warn!(
+                "Optimistic block in is not updated or optimistic block less than final block"
+            );
+            crate::metrics::OPTIMISTIC_UPDATING.set_not_working();
+        };
+
+        // When optimistic block is updated and optimistic block more than final block
```
khorolets, line 293:
```suggestion
        // When an optimistic block is updated, and it is greater than the final block
```

khorolets (changes requested):

khorolets (approved):

PR #240, opened 2024-04-25
add fetch protocol config for epoch

kobayurii:


configuration/src/configs/general.rs
```diff
@@ -277,9 +274,8 @@ impl From<CommonGeneralConfig> for GeneralRpcServerConfig {
             )
             .expect("Invalid redis url"),
             referer_header_value: common_config
-                .rpc_server
                 .referer_header_value
-                .unwrap_or_else(CommonGeneralRpcServerConfig::default_referer_header_value),
+                .unwrap_or("read-rpc".to_string()),
```
khorolets, line 278: Consider moving the default value into a const, make it `http://read-rpc.local` to give a good example for others (referer should be a valid URL)

khorolets (approved):

Issue #236, opened 2024-04-24
Performance problem ft. `near-vm`


khorolets:
After upgrading the underlying dependency `near-vm` (to the 1.39.x nearcore) that we use to support `query.call_function` (view calls), we observe the performance issue.

The ReadRPC node eats up all the RAM under the load (250-300 req/s) in about 17 minutes. We checked and compared it with the previous version and figured it still dies, but later (load 500-800 req/s / ~12h).

We are currently narrowing down the problem and trying to find a solution. Current assumptions:

1. We use `near-vm` in the wrong way
1. Custom-ish caches we pass to `near-vm` has a bug
1. `near-vm` has a bug

We bet on the first one and try to figure out the correct way of using it and what we are doing wrong.

### Some tech details about what we are checking

We invoke `near-vm` [here](https://github.com/near/read-rpc/blob/0b7a59cbf691f18751d640e5907a77a14f8b4537/rpc-server/src/modules/queries/utils.rs#L106-L147) using `tokio::task::spawn_blocking()` since `near_vm_runner::run` is synchronous.

*We tried to remove the `spawn_blocking`, and it helped, but the throughput dropped to about 20 requests per second. Not an option for us*

So, right now, we are trying to optimize that call to achieve the best performance and resource consumption.

khorolets:
I have a lead.

When we call `with_compiled_and_loaded` method under the hood it accepts the `code: Option<&ContractCode>` 

ref: https://github.com/near/nearcore/blob/14334b9e4e2bff198537130324f2bb9702f54d6d/runtime/near-vm-runner/src/near_vm_runner.rs#L348-L359

And following down the road, it appears that if this value is `None`, it is a signal for the `near-vm-runner` to compile the contract even if we have it in the cache, and it hasn't changed ever since.

That looks like a place where we use the `near-vm-runner` crate incorrectly 🤦‍♂️ We naively pass the code every time, hoping `near-vm-runner` handles it automatically, but it appears it does not.

I've hacked a bit and checked on one contract with our synthetic test and `heaptrack` profiling shows we don't leak.

@kobayurii will implement logic to decide whether to pass the code using the `run` method. After that, we will be able to stress test it to ensure it helped.

kobayurii:
https://github.com/near/read-rpc/pull/245

khorolets:
The performance issue still exists.

[...]

PR #241, opened 2024-04-29
(rpc_server): Performance problem ft. near-vm

kobayurii:
When we call `near_vm_runner::run` method accepts the code: Option<&ContractCode>

ref: https://github.com/near/nearcore/blob/14334b9e4e2bff198537130324f2bb9702f54d6d/runtime/near-vm-runner/src/runner.rs#L46

And following down the road, it appears that if this value is None, it is a signal for the near-vm-runner to compile the contract even if we have it in the cache, and it hasn't changed ever since.

After updating nearcore to 1.39.x we use the near-vm-runner crate incorrectly 🤦‍♂️ We naively pass the code every time, hoping near-vm-runner handles it automatically, but it appears it does not.

### DIFF
--- a/rpc-server/src/metrics.rs
+++ b/rpc-server/src/metrics.rs
@@ -665,18 +665,6 @@ lazy_static! {
     ).unwrap();
 }
 
-lazy_static! {
-    // Execute contract counters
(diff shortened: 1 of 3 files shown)


kobayurii:
https://github.com/near/read-rpc/issues/236

khorolets (approved):

PR #247, opened 2024-05-04
feat: Store `TransactionDetails` in object storage (GCS) 

khorolets:
## Rational 

We noticed that the `transaction_details` table in ScyllaDB takes ~2 TB of data, directly affecting the infra cost. We want to reduce the cost even more. Building on our successful experience of serving some data directly from the NEAR Lake AWS S3 storage, we propose to move the `TransactionDetails` data to object storage (GSC) to further offload ScyllaDB.

Since this structure is stored in bytes (borsh-serialized), it doesn't make any big difference to the performance (though some increase in latency, anyway, is expected, not that significant to overpay).

### What is done

1. I've introduced a small crate (lib) `tx-details-storage` that interacts with the provided S3-compatible object storage to store and retrieve raw bytes
2. Extended the `configuration` crate with the dedicated config section for `tx-details-storage`. Updated `config.example.toml` to reflect new accepted parameters
3. Refactored `tx-indexer` to store the finished `TransactionDetails` to the object storage using the `tx-details-storage` library. **The library doesn't handle serialization/deserialization to keep it as simple as possible**, this is left for the indexer and rpc-server.
4. Refactored `rpc-server` to retrieve `TransactionDetails` from the object storage using the newly introduced library.
5. Additionally, I've extended the `tx-indexer` with a few metrics to monitor what's happening there.

**Important note:** Recently, we've adjusted the `rpc-server` to return half-baked (not finished, in progress) transaction details from the database cache table. This logic is preserved.

### Next steps

We must wait to deliver this change. We have yet to get the data in an object storage.

I plan to make it in three phases:

1. ✅ Create a small script that will walk over each transaction present in NEAR Protocol and copy the `TransactionDetails` from ScyllaDB to GCS. **This has to be done for both testnet and mainnet**
2. Start the new `tx-indexer` (from this PR) to continue collecting the data into the object storage
3. Replace the `rpc-server` instances with the new ones that can read from the object storage
4. (cleanup phase) Stop old `tx-indexer`s, drop `transaction_details` table from ScyllaDB, downscale ScyllaDB

### Update from 2024-07-03

The table `transaction_details` is growing and reaching the limits of the database we have right now. We don't want to scale it, so we need to stop the growth in the short-term. I refactored code a bit by leaving the legacy way of searching for the transaction in the database (Scylla) while we migrate.

I've added an additional metric for this `legacy_database_tx_details` to monitor the migration period. We expect that counters to be null for some time before we can consider migration as finished.

### DIFF
--- a/tx-indexer/src/collector.rs
+++ b/tx-indexer/src/collector.rs
@@ -1,3 +1,4 @@
+use borsh::BorshDeserialize;
 use futures::{
     future::{join_all, try_join_all},
     StreamExt,
@@ -21,6 +22,7 @@ pub(crate) async fn index_transactions(
     streamer_message: &near_indexer_primitives::StreamerMessage,
     db_manager: &std::sync::Arc<Box<dyn database::TxIndexerDbManager + Sync + Send + 'static>>,
     tx_collecting_storage: &std::sync::Arc<crate::storage::HashStorageWithDB>,
+    tx_details_storage: &std::sync::Arc<crate::TxDetailsStorage>,
     indexer_config: &configuration::TxIndexerConfig,
 ) -> anyhow::Result<()> {
     extract_transactions_to_collect(
@@ -42,10 +44,12 @@ pub(crate) async fn index_transactions(
 
     if !finished_transaction_details.is_empty() {
         let db_manager = db_manager.clone();
+        let tx_details_storage = tx_details_storage.clone();
         tokio::spawn(async move {
-            let send_finished_transaction_details_futures = finished_transaction_details
-                .into_iter()
-                .map(|tx_details| save_transaction_details(&db_manager, tx_details));
+            let send_finished_transaction_details_futures =
+                finished_transaction_details.into_iter().map(|tx_details| {
+                    save_transaction_details(&db_manager, &tx_details_storage, tx_details)
+                });
 
             join_all(send_finished_transaction_details_futures).await;
         });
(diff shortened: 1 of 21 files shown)


tx-indexer/src/collector.rs
```diff
@@ -292,81 +297,83 @@ async fn save_transaction_details(
...
-                tracing::warn!(
+    // We faced the issue when the transaction was saved to the storage but later
+    // was failing to deserialize. To avoid this issue and monitor the situation
+    // without runing the user experience, we will try to save the transaction
+    // to the storage and validate that it is saved correctly for 3 attempts
+    // before we throw an error.
+    let mut save_attempts = 0;
+    'retry: loop {
+        save_attempts += 1;
+        if save_attempts > 3 {
```
kobayurii, line 310: Let's increase the number of attempts to 10 to make sure we've done everything we can to save the transaction.

tx-indexer/src/collector.rs
```diff
@@ -292,81 +297,83 @@ async fn save_transaction_details(
...
+                    tokio::time::sleep(std::time::Duration::from_millis(150)).await;
+                    let Ok(tx_details_bytes_from_storage) =
+                        tx_details_storage.retrieve(&transaction_hash).await
+                    else {
+                        tracing::error!(
+                            target: crate::INDEXER,
+                            "Failed to retrieve transaction {} from storage",
+                            transaction_hash,
+                        );
+                        continue 'validator;
```
kobayurii, line 336: I think we need to add a separate counter for this loop in case we hit infinity for some unknown reason, then we can get out of it. What do you think?

kobayurii (changes requested):
Looks great. Thank you! Just a couple places that caused me concern. Let's solve them and we are good to merge this.

PR #251, opened 2024-05-17
Metrics improvement

kobayurii:


kobayurii:
https://github.com/near/read-rpc/pull/255

PR #254, opened 2024-05-17
fix(near-state-indexer): Fix the TypeError in communication with Redis leading to spam of warnings

khorolets:
The issue is relatively small and the only outcome of it was a bunch of warning, the logic was intact.


khorolets:
I've built the `near-state-indexer` binary for production and it is already deployed with this fix included. I don't thing we need to rush with the additional release. The problem is relatively small, we will land it fix to the release along with 1.40.x nearcore

kobayurii (approved):

PR #258, opened 2024-05-22
refactor(rpc-server): Tweaks and clean up around query.call_function method implementation method implementation

khorolets:
## 🚧 WIP

I am thoroughly reviewing the logic we have around `query.call_function` method implementation to find a way to improve the usage of ScyllaDB here. This work is related to the #236 

While I am here and noticing some areas for improvement I am implementing it as well (Boy Scout Rule in action).

### High-level changes
* Extract the logic of contract execution to a separate module `contract_runner` within the `query` one
* Small improvement for avoiding the clone of `block_reference` for `fetch_block_from_cache_or_get` function (this actually went a bit beyond the `query` module tbh)
* Simplified the logic of optimistic/final data fetch in `query:: function_call` (dropped two similar functions, moved the code to the `function_call` body, it seems like it belongs there to me)


rpc-server/src/modules/queries/contract_runner/code_storage.rs
```diff
@@ -0,0 +1,305 @@
...
+pub type Result<T> = ::std::result::Result<T, near_vm_runner::logic::VMLogicError>;
+
+pub struct CodeStorage {
+    db_manager: std::sync::Arc<Box<dyn ReaderDbManager + Sync + Send + 'static>>,
+    account_id: near_primitives::types::AccountId,
+    block_height: near_primitives::types::BlockHeight,
+    validators: HashMap<near_primitives::types::AccountId, near_primitives::types::Balance>,
+    data_count: u64,
+    is_optimistic: bool,
+    prefetched_data:
```
kobayurii, line 16: This is already deleted in https://github.com/near/read-rpc/pull/259. Observations have shown that it is not efficient to request the whole stack. Therefore, this functionality was removed.

rpc-server/src/modules/queries/contract_runner/code_storage.rs
```diff
@@ -0,0 +1,305 @@
...
+            validators,
+            data_count: Default::default(), // TODO: Using for generate_data_id
+            is_optimistic: !optimistic_data.is_empty(),
+            prefetched_data,
+            optimistic_data,
+        }
+    }
+
+    fn get_state_key_data(&self, key: &[u8]) -> readnode_primitives::StateValue {
+        if let Some(prefetched_data) = &self.prefetched_data {
```
kobayurii, line 63: It is not needed according to the comment above.

rpc-server/src/modules/queries/contract_runner/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+                    .put(contract.data.code_hash(), code.data.clone())
+                    .await;
+                near_vm_runner::ContractCode::new(code.data, Some(contract.data.code_hash()))
+            }
+        })
+    };
+
+    // TODO: Refactor this part. It's a temporary solution to fetch state keys from DB for the poolv1.near contracts
+    // https://github.com/near/read-rpc/issues/150
+    let contract_state = if account_id.to_string().ends_with("poolv1.near") {
```
kobayurii, line 116: It is not needed according to the comment above.

kobayurii (approved):
This is a great improvement! Thank you!

PR #259, opened 2024-05-22
(database): Implement metrics to count requests for each table

kobayurii:
This pull request introduces new metrics to track the number of requests made to each database table. By implementing these metrics, we can monitor the usage patterns and access frequency of individual tables, providing valuable insights for performance optimization and capacity planning.


khorolets (approved):
Great! Thank you!

PR #266, opened 2024-05-31
feat: added `view_receipt_record` rpc endpoint

### LINKED ISSUE #264: Add a JSON-RPC method to get a parent transaction hash for a given receipt ID
## Goal

We need a new method accepting `receipt_id: String` that will respond with either "not found" or some set of data about where the given receipt originated from:

Possible structure:

```
{
  "receipt_id", // Requested receipt id for consistency
  "parent_transaction_hash" // the hash of the transaction that initiated the events leading to the requested receipt creation
  "block_height" // height where the receipt was observed
  "shard_id" // ID of the shard the receipt was observed
}
```

### Context

Matching receipts with the transaction they originated from is a long-standing issue for many developers who work closely with data. For all these years, people have invented different workarounds to handle this problem. When we started the ReadRPC project, we made the `receipts_map` table and collected data.

```
"CREATE TABLE IF NOT EXISTS receipts_map (
    receipt_id varchar,
    block_height varint,
    parent_transaction_hash varchar,
    shard_id varint,
    PRIMARY KEY (receipt_id)
)
```

Implementing this method is low-hanging fruit. It is not tough to implement but introduces a lot of value for the community.

## Helpful references

* @kobayurii has introduced a method native JSON-RPC doesn't have [`view_state_paginate`](https://github.com/near/read-rpc/blob/develop/rpc-server/src/modules/state/methods.rs#L7) might be a good starting point for inspiration
* Despite the approach of naming a new method with the prefix `EXPERIMENTAL_` on the native JSON-RPC, we decided not to follow

### Testing data

ScyllaDB doesn't provide any way to back up the data, and the feature might be needed for testing. ~~I am~~ @kobayurii is attaching the CSV with the data. **We don't have a tool to propagate the data to the development setup from CSV. It's up to a developer to handle this. Sorry.**

// Sample data
[receipts_map.csv](https://github.com/near/read-rpc/files/15494839/receipts_map.csv)

frolvanya:
Here's an example of calling `view_receipt_record`:
image

closes #264

### DIFF
--- a/rpc-server/src/modules/receipts/methods.rs
+++ b/rpc-server/src/modules/receipts/methods.rs
@@ -40,9 +40,8 @@ pub async fn view_receipt_record(
     tracing::debug!("`view_receipt_record` call. Params: {:?}", params);
     let receipt_request =
         near_jsonrpc::primitives::types::receipts::RpcReceiptRequest::parse(params)?;
-    crate::metrics::RECEIPT_RECORD_REQUESTS_TOTAL.inc();
 
-    let result = fetch_receipt_record(&data, &receipt_request).await;
+    let result = fetch_receipt_record(&data, &receipt_request, "view_receipt_record").await;
 
     Ok(result.map_err(near_jsonrpc::primitives::errors::RpcError::from)?)
 }
@@ -57,7 +56,7 @@ async fn fetch_receipt(
 > {
     let receipt_id = request.receipt_reference.receipt_id;
 
-    let receipt_record = fetch_receipt_record(data, request).await?;
+    let receipt_record = fetch_receipt_record(data, request, "EXPERIMENTAL_receipt").await?;
 
     // Getting the raw Vec<u8> of the TransactionDetails from ScyllaDB
     let transaction_details = data
@@ -97,6 +96,7 @@ async fn fetch_receipt(
 async fn fetch_receipt_record(
     data: &Data<ServerContext>,
     request: &near_jsonrpc::primitives::types::receipts::RpcReceiptRequest,
+    method_name: &str,
 ) -> Result<
     readnode_primitives::ReceiptRecord,
     near_jsonrpc::primitives::types::receipts::RpcReceiptError,
@@ -104,10 +104,10 @@ async fn fetch_receipt_record(
     let receipt_id = request.receipt_reference.receipt_id;
 
     data.db_manager
-        .get_receipt_by_id(receipt_id)
+        .get_receipt_by_id(receipt_id, method_name)
         .await
         .map_err(|err| {
-            tracing::warn!("Error in `fetch_receipt_record` call: {:?}", err);
+            tracing::warn!("Error in `{}` call: {:?}", method_name, err);
             near_jsonrpc::primitives::types::receipts::RpcReceiptError::UnknownReceipt {
                 receipt_id,
             }
--- a/rpc-server/src/modules/receipts/methods.rs
+++ b/rpc-server/src/modules/receipts/methods.rs
@@ -36,14 +36,16 @@ pub async fn receipt(
 pub async fn view_receipt_record(
     data: Data<ServerContext>,
     Params(params): Params<serde_json::Value>,
-) -> Result<readnode_primitives::ReceiptRecord, RPCError> {
+) -> Result<crate::modules::receipts::RpcReceiptRecordResponse, RPCError> {
     tracing::debug!("`view_receipt_record` call. Params: {:?}", params);
     let receipt_request =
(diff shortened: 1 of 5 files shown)


rpc-server/src/modules/receipts/methods.rs
```diff
@@ -36,6 +36,38 @@ pub async fn receipt(
...
+        near_jsonrpc::primitives::types::receipts::RpcReceiptRequest::parse(params)?;
+    crate::metrics::RECEIPT_RECORD_REQUESTS_TOTAL.inc();
+
+    #[cfg(feature = "shadow_data_consistency")]
+    {
+        let shadow_result = crate::utils::shadow_compare_results_handler(
+            crate::metrics::RECEIPT_RECORD_REQUESTS_TOTAL.get(),
+            data.shadow_data_consistency_rate,
+            &fetch_receipt_record(&data, &receipt_request).await,
+            data.near_rpc_client.clone(),
```
khorolets, line 56: The entire shadow data consistency check is redundant here. There is no such method on native RPC, so there is nothing to compare. Please drop it.

rpc-server/src/modules/receipts/methods.rs
```diff
@@ -87,3 +110,25 @@ async fn fetch_receipt(
...
+    readnode_primitives::ReceiptRecord,
+    near_jsonrpc::primitives::types::receipts::RpcReceiptError,
+> {
+    let receipt_id = request.receipt_reference.receipt_id;
+
+    data.db_manager
+        .get_receipt_by_id(receipt_id)
+        .await
+        .map_err(|err| {
+            println!("Error in `fetch_receipt_record` call: {:?}", err);
```
khorolets, line 128: leftover print

rpc-server/src/metrics.rs
```diff
@@ -166,6 +166,10 @@ lazy_static! {
         "receipt_requests_counter",
         "Total number requests to the receipt endpoint"
     ).unwrap();
+    pub static ref RECEIPT_RECORD_REQUESTS_TOTAL: IntCounter = try_create_int_counter(
```
khorolets, line 169: We have migrated to the common total counter with labels, I believe @kobayurii made a middleware for it.

I will let him to elaborate on that

khorolets (changes requested):
Nice job indeed! I left a few change requests, but please wait for @kobayurii review before rushing on it. Thank!

frolvanya:
Thanks for a review! I guess I need to pay more attention to delete all debug leftovers. I've fixed first two issues and I'll wait for @kobayurii review, before pushing changes. However, I still don't know if it's fine to return non Response type from RPC endpoint. Can someone tell me if it's a good practice:
https://github.com/near/read-rpc/blob/ac8f4f221afe391c20144cd7c7397868d13a525c/readnode-primitives/src/lib.rs#L180-L181
https://github.com/near/read-rpc/blob/ac8f4f221afe391c20144cd7c7397868d13a525c/rpc-server/src/modules/receipts/methods.rs#L39-L44

kobayurii (changes requested):
Thank you! Great job! I left a little comment. Also, I noticed that you created a pull request in the main branch.  Please reopen the pull request for the develop branch. `develop` is the main branch for development. Also don't forget to fix the checker.

khorolets:
> However, I still don't know if it's fine to return non Response type from RPC endpoint. Can someone tell me if it's a good practice:
> 
> https://github.com/near/read-rpc/blob/ac8f4f221afe391c20144cd7c7397868d13a525c/readnode-primitives/src/lib.rs#L180-L181
> 
> 
> https://github.com/near/read-rpc/blob/ac8f4f221afe391c20144cd7c7397868d13a525c/rpc-server/src/modules/receipts/methods.rs#L39-L44

While it's not perfect either, I would suggest following what @kobayurii did for the other custom method we have here 

https://github.com/near/read-rpc/blob/develop/rpc-server/src/modules/state/mod.rs

I am not a fan of duplicating structures, but I would ask you to have the `Resuest/Response` structures defined in the `modules/receipts/mod.rs`

Feel free to implement `From<ReceiptRecord>` or `TryFrom` for your request/response structures. And drop the serde derives from the original structure.

My thoughts come from the idea of having dedicated structures for (again) request/response. Having `From` or `TryFrom` from the internal structure would alert us if we change something in the future.

Thank you for drawing my attention to this question 🙏

khorolets:
@frolvanya and please update https://github.com/near/read-rpc/blob/main/docs/CUSTOM_RPC_METHODS.md with info about the new method you've added. Thank you.

khorolets (approved):
Perfect! Thank you 🙏

PR #272, opened 2024-06-06
(rpc_server): Extend metrics

kobayurii:
In this PR Historical/regular counter is extended and broken down by method.
This metric should to help to analyze the block_references for methods 
to see the most common pattern is it specific block or final/optimistic.

- final
- optimistic
- block_id:
     - regular - includes 5 last epochs
     - historical - later than 5 epochs ago

### DIFF
--- a/database/src/base/rpc_server.rs
+++ b/database/src/base/rpc_server.rs
@@ -90,18 +90,21 @@ pub trait ReaderDbManager {
     ) -> anyhow::Result<readnode_primitives::ReceiptRecord>;
 
     /// Returns the readnode_primitives::TransactionDetails at the given transaction hash
+    /// TODO: rewrite this method to return a Result like a struct with block_height and transaction_details instead of a tuple
     async fn get_transaction_by_hash(
         &self,
         transaction_hash: &str,
         method_name: &str,
     ) -> anyhow::Result<(u64, readnode_primitives::TransactionDetails)>;
     /// Returns the readnode_primitives::TransactionDetails at the given transaction hash
+    /// TODO: rewrite this method to return a Result like a struct with block_height and transaction_details instead of a tuple
     async fn get_indexed_transaction_by_hash(
         &self,
         transaction_hash: &str,
         method_name: &str,
     ) -> anyhow::Result<(u64, readnode_primitives::TransactionDetails)>;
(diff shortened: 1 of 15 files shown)


rpc-server/src/metrics.rs
```diff
@@ -96,12 +96,18 @@ lazy_static! {
...
 
-    pub(crate) static ref REQUESTS_COUNTER: IntCounterVec = register_int_counter_vec(
-        "requests_counter",
+    pub(crate) static ref TOTAL_REQUESTS_COUNTER: IntCounterVec = register_int_counter_vec(
+        "total_requests_counter",
         "Total number of requests",
         &["request_type"] // This declares a label named `request_type`
     ).unwrap();
 
+    pub(crate) static ref TIMELINE_METHOD_REQUESTS_COUNTER: IntCounterVec = register_int_counter_vec(
```
khorolets, line 105: I think it'd be better to merge these two metrics `TOTAL_REQUESTS_COUNTER` and `TIMELINE_METHOD_REQUESTS_COUNTER` into one.

My assumption is that using Grafana we will be able to sum up the broken down metrics into a single value (if needed).

What do you think? Is there any benefit of having two similar yet different in details metrics?

database/src/base/rpc_server.rs
```diff
@@ -94,19 +94,19 @@ pub trait ReaderDbManager {
         &self,
         transaction_hash: &str,
         method_name: &str,
-    ) -> anyhow::Result<readnode_primitives::TransactionDetails>;
+    ) -> anyhow::Result<(u64, readnode_primitives::TransactionDetails)>;
```
khorolets, line 97: I agree, sometimes there might be a situation where returning a tuple is inevitable. Although, I don't believe this is such a situation.

Please consider introducing something like

```rust
struct ResponseWithBlockHeight<T> {
  block_height: u64,
  response: T,
}
```

Thus it'll be more maintainable in the future and less error-prone.

khorolets (changes requested):
Overall, the metric improvements look good. I've highlighted a few places I think we'd benefit from.

database/src/base/rpc_server.rs
kobayurii: We expected to rewrite the database crate. I left a TODO comment for the future.

khorolets (approved):

PR #278, opened 2024-06-10
fix: clarified Redis connection errors and added Redis dependency to docker-compose files

### LINKED ISSUE #274: (rpc-server): Improve error handling around "Connection refused"
To support optimistic block finalities queries we've introduced a dependency on Redis. A dedicated node of `near-state-indexer` fills it with the data of final and optimistic blocks.

We haven't documented it properly and didn't update the `docker-compose.yml` we have in the project root. Also, the error we observe if we start `rpc-server` without a connection to Redis is too ambiguous "Connection refused ..." 

On the other hand, for the `rpc-server` to work the connection to Redis just need to be established and if it is not filled with the data the server continues to work. There is a fallback mechanism implemented. That's why I wonder if it makes sense to give a meaningful warning but handle the `rpc-server` start initialization step to not fail if the Redis connection is absent? 

We need to make this chores to keep the repo in a good shape:

* [ ] Improve the error to make it less ambiguous, we need a clear message identifying it is related to connection to Redis
* [ ] Update the `docker-compose.yml` to ensure it is always in a working condition to `docker-compose up` from `main`
* [ ] #279

### LINKED ISSUE #279: (rpc-server): Make Redis dependency optional
Continuation of #274 

First of all, we need to decide how to proceed with having Redis as a mandatory dependency and whether the `rpc-server` can work without it even with limited capabilities.

From what I can gather, we have all the logic to fallback into alternative way of getting the data if the Redis is not available. While it is not something we want in production environment, it is totally acceptable in the development.

I need to hear if @kobayurii has second thoughts about it and we can proceed with this issue.

frolvanya:
- [x] Improve the error to make it less ambiguous, we need a clear message identifying it is related to connection to Redis

  Before: 
  image

  After:
  image

- [x] Update the `docker-compose.yml` to ensure it is always in a working condition to `docker-compose up` from `main`

  To complete this, I've added redis dependency to postgres and rightsizing docker-compose files

- [x] Decide how to proceed with having Redis as a mandatory dependency and whether the `rpc-server` can work without it even with limited capabilities

  In any case we can merge these minor fixes now and later (@kobayurii @khorolets) will decide wether Redis is required or not

closes #274 
closes #279

### DIFF
--- a/examples/postgres-compose.yml
+++ b/examples/postgres-compose.yml
@@ -2,6 +2,7 @@ version: '3.7'
 
 x-common-variables: &common-variables
   CHAIN_ID: mainnet
+  REDIS_URL: redis://redis:6379/
   DATABASE_URL: postgres:5432
   DATABASE_USER: postgres
   DATABASE_PASSWORD: password
@@ -29,6 +30,7 @@ services:
       - 8000:8000
     depends_on:
       - postgres
+      - redis
     restart: on-failure
 
   state-indexer:
@@ -45,6 +47,7 @@ services:
       - 8080:8080
     depends_on:
       - postgres
+      - redis
     restart: on-failure
 
   tx-indexer:
@@ -61,6 +64,7 @@ services:
       - 8081:8081
     depends_on:
       - postgres
+      - redis
     restart: on-failure
 
   postgres:
@@ -72,3 +76,9 @@ services:
       POSTGRES_PASSWORD: password
     ports:
       - 5432:5432
+
+  redis:
+    image: redis/redis-stack-server:latest
+    restart: always
+    ports:
+      - 6379:6379
--- a/examples/rightsizing-compose.yml
+++ b/examples/rightsizing-compose.yml
@@ -2,6 +2,7 @@ version: '3.7'
 
 x-common-variables: &common-variables
   CHAIN_ID: mainnet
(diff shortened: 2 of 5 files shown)


khorolets (approved):
Looks great! I will follow up in a separate issue about decision to transform the Redis dependency to optional

kobayurii (changes requested):
Hello! Thank you! I think we should use radis for development. It will only be used if we have a running instance of near_state_indexer which is based on nearcore. I would like you to check at the initialization stage whether the connection to redis is established and if there is an error then do not spawn the tasks `update_final_block_regularly_from_redis` and `update_optimistic_block_regularly` and `OPTIMISTIC_UPDATING.set_not_working()`
and generated a Warning error. 
This way you won't have to carry dependencies in all places.

frolvanya:
> I would also like you to update the `readme.md` specify redis as an optional dependency and describe its use only when paired with near_state_indexer.

Should it be done in main README.md file or inside the rpc-server/README.md?

khorolets:
> Should it be done in root `README.md` file or inside the `rpc-server/README.md`?

It's enough for now, we'll perform docs cleaning later.

kobayurii (approved):
Looks good! Thank you!

PR #280, opened 2024-06-20
chore: Upgrade to nearcore 1.40.0

khorolets:
WIP: there are some breaking changes around `near-vm-runner` we need to adapt

* Upgrade dependencies for supporting `nearcore-1.40.0`


kobayurii (approved):
Grate! Thank you!

PR #285, opened 2024-06-26
(database-new): optimize state-indexer shard layout respect

khorolets:
Some code optimization around `state-indexer` and futures we use there to fill the data


kobayurii (approved):
Great! Thank you!

PR #294, opened 2024-07-03
fix: rpc-server tx fallback to legacy improve, tx-indexer adjust timeouts and log levels

khorolets:
* (read-rpc) Fall back to legacy database if tx failed to read from bucket (deserialization error, or not found)
* (read-rpc) Change the method of getting the tx for `EXPERIMENTAL_receipt`
* (tx-indexer) Adjust timeouts for saving and validating the tx_details, adjust log levels


kobayurii (approved):
Great! Thank you!

PR #300, opened 2024-07-08
feat: prefetch optimization

frolvanya:
This PR introduces a configuration option to control the maximum size of storage used for state prefetching during a `view_call`. Prefetching helps to speed up the `view_call` by loading data in advance, but it can consume more time if state size is too big. To prevent excessive time usage during a `view_call`, we set a default limit of 1MB (1_000_000 bytes) for the prefetch storage. Prefetching will only be an option if the state storage usage is below this given threshold

### DIFF
--- a/rpc-server/src/modules/queries/contract_runner/code_storage.rs
+++ b/rpc-server/src/modules/queries/contract_runner/code_storage.rs
@@ -46,24 +46,20 @@ impl CodeStorage {
             readnode_primitives::StateKey,
             Option<readnode_primitives::StateValue>,
         >,
-        prefetch_storage_limit: u64,
+        storage_usage: u64,
+        prefetch_state_size_limit: u64,
     ) -> Self {
         let mut prefetch_state_data = HashMap::new();
 
-        if let Ok(account) = db_manager
-            .get_account(&account_id, block_height, "query_call_function")
-            .await
-        {
-            if account.data.storage_usage() < prefetch_storage_limit {
-                prefetch_state_data = utils::get_state_from_db(
-                    &db_manager,
-                    &account_id,
-                    block_height,
-                    &[],
-                    "query_call_function",
-                )
-                .await;
-            }
+        if storage_usage < prefetch_state_size_limit {
+            prefetch_state_data = utils::get_state_from_db(
+                &db_manager,
+                &account_id,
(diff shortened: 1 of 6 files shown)


kobayurii (changes requested):
Overall, looks good. I've highlighted a few places I think could be improved. Also, please leave more description in the PR about why we are doing this. This may help us in the future.

frolvanya:
I've addressed all issues and changed a PR description, but I'm not sure if new docs are more explanatory or I need to improve them more

rpc-server/src/modules/queries/contract_runner/code_storage.rs
```diff
@@ -46,19 +46,21 @@ impl CodeStorage {
             readnode_primitives::StateKey,
             Option<readnode_primitives::StateValue>,
         >,
+        storage_usage: u64,
+        prefetch_state_size_limit: u64,
     ) -> Self {
-        let prefetch_state_data = if account_id.to_string().ends_with("poolv1.near") {
-            utils::get_state_from_db(
+        let mut prefetch_state_data = HashMap::new();
```
kobayurii, line 52: I don't think it should be `mut`.

rpc-server/src/modules/queries/contract_runner/code_storage.rs
```diff
@@ -46,19 +46,21 @@ impl CodeStorage {
...
+
+        if storage_usage < prefetch_state_size_limit {
```
kobayurii, line 54:
```rust
let prefetch_state_data = if storage_usage < prefetch_state_size_limit {
     utils::get_state_from_db(
        &db_manager,
        &account_id,
        block_height,
        &[],
        "query_call_function",
    )
    .await
} else {
    HashMap::new()
}

kobayurii (changes requested):

rpc-server/src/modules/queries/contract_runner/code_storage.rs
frolvanya: Good catch, thanks! I forgot to change this after removing second request for account info

kobayurii (approved):
Great! Thank you!

khorolets (approved):

PR #301, opened 2024-07-10
feat: added support for a `version` flag

### LINKED ISSUE #287: (indexers): Ensure all the indexers respond to `--version`
### Motivation

Recently, I've upgraded the `near-state-indexer` to the newest `nearcore-1.40.0`, made the ReadRPC release and all. And after the deploy I couldn't prove the version of the `near-state-indexer` binary is the newer, because we don't support the `--version` for some reason.

### TODO:

* Make sure every indexer within the project responds to the `--version` with the most meaningful information. 
* [Ideally] the version should include both the project version (ReadRPC) and the supported `nearcore` version.

frolvanya:
I've added the version flag for all indexers. However, I didn't complete the part with a supported nearcore version. In order to print an information about nearcore version along with the indexer version, we need to store nearcore version in some env variable (or in .toml file). I'm not sure that this overhead is needed, but I'm more than happy to implement it, so please say if this small overhead worth it or not

closes #287

### DIFF
--- a/logic-state-indexer/src/configs.rs
+++ b/logic-state-indexer/src/configs.rs
@@ -3,7 +3,7 @@ pub use clap::{Parser, Subcommand};
 /// NEAR Indexer for Explorer
 /// Watches for stream of blocks from the chain
 #[derive(Parser, Debug)]
-#[command(version)]
+#[command(version = concat!(env!("CARGO_PKG_VERSION"), "\nnearcore ", env!("NEARCORE_VERSION")))]
 pub struct Opts {
     #[clap(subcommand)]
     pub start_options: StartOptions,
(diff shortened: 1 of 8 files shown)


kobayurii (changes requested):
Thank you! I left you a comment where you can find the information you need.

You can find nearcore version in the .cargo/config.toml -> NEARCORE_VERSION
Usage example: near-state-indexer/build.rs

frolvanya:
Thanks for the info! I've added info about nearcore into the version flag. Let me know if the format is strange and should be changed. It looks like this now:
image

kobayurii (approved):
Greate! Thank you!

frolvanya:
@khorolets I've added info about pkg version to the logs to all indexers

image

PR #303, opened 2024-07-12
fix: view state unknown account error

### LINKED ISSUE #302: (rpc-server): Result mismatch for an account with no keys in state
@evgenykuzyakov reported:

Found inconsistency between the read-rpc and a regular rpc. When you query ViewState for an account without any keys in the state, it returns HandlerError(UnknownAccount while regular RPC returns empty array of keys

---

I suspect this is related to the way we handle state changes. If the account never had any keys in a state we assume such account does not exist. Since we don't keep a list of accounts it might be a tricky task to solve

frolvanya:
This PR closes #302. Basically current fix is very straightforward, therefore it might not be very optimal. We simply check whether an account exists in our DB and handle "Unknown account" error before fetching the state. If we didn't exit function during the check, then we print the state no matter of its value (even if the state is empty). I also thought about the approach that would require to change the raw sql statement of getting the state from DB, maybe by utilizing join command on `state_changes_account` and `state_changes_data`, but this would be also an extra call to db and it will be quite heavy, so I decided to stick to this approach. I guess that in order to come up with the most optimal solution we need to change the way how we store `state_changes_data`

Before:
image

After:
image

### DIFF
--- a/rpc-server/src/modules/queries/methods.rs
+++ b/rpc-server/src/modules/queries/methods.rs
@@ -401,6 +401,17 @@ async fn view_state(
         is_optimistic,
     );
 
+    data.db_manager
+        .get_account(account_id, block.block_height, "query_view_state")
+        .await
+        .map_err(
+            |_err| near_jsonrpc::primitives::types::query::RpcQueryError::UnknownAccount {
+                requested_account_id: account_id.clone(),
+                block_height: block.block_height,
+                block_hash: block.block_hash,
+            },
+        )?;
+
     let state_item = if is_optimistic {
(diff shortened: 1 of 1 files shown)


frolvanya:
It might be also a good idea to move this check directly to the `view_state` function to remove code duplication

https://github.com/near/read-rpc/blob/133110b655c448932b5a9e49db24e79dbcf1dbf6/rpc-server/src/modules/queries/methods.rs#L432-L441

khorolets (approved):
Great! Kudos @frolvanya for proactivity!

PR #310, opened 2024-07-16
feat: added config checks

frolvanya:
This PR introduces additional validation checks for certain configuration variables. Specifically, it ensures that all URLs and `f64` values adhere to their expected formats and constraints

Here's an example of running rpc-server with invalid env variables:
image

### DIFF
--- a/configuration/src/configs/database.rs
+++ b/configuration/src/configs/database.rs
@@ -1,15 +1,18 @@
-use crate::configs::{deserialize_data_or_env, deserialize_url_or_env};
+use crate::configs::deserialize_data_or_env;
+use validator::Validate;
+
 use near_lake_framework::near_indexer_primitives::near_primitives;
 
 // Database connection URL
 // Example: "postgres://user:password@localhost:5432/dbname"
 type DatabaseConnectUrl = String;
 
-#[derive(serde_derive::Deserialize, Debug, Clone, Default)]
+#[derive(Validate, serde_derive::Deserialize, Debug, Clone, Default)]
 pub struct ShardDatabaseConfig {
     #[serde(deserialize_with = "deserialize_data_or_env")]
     pub shard_id: u64,
-    #[serde(deserialize_with = "deserialize_url_or_env")]
+    #[validate(url(message = "Invalid database shard URL"))]
+    #[serde(deserialize_with = "deserialize_data_or_env")]
     pub database_url: DatabaseConnectUrl,
 }
 
@@ -33,10 +36,12 @@ impl DatabaseConfig {
     }
 }
 
-#[derive(serde_derive::Deserialize, Debug, Clone, Default)]
+#[derive(Validate, serde_derive::Deserialize, Debug, Clone, Default)]
 pub struct CommonDatabaseConfig {
-    #[serde(deserialize_with = "deserialize_url_or_env")]
+    #[validate(url(message = "Invalid database URL"))]
+    #[serde(deserialize_with = "deserialize_data_or_env")]
     pub database_url: DatabaseConnectUrl,
+    #[validate(nested)]
     #[serde(default)]
     pub shards: Vec<ShardDatabaseConfig>,
 }
--- a/configuration/src/configs/database.rs
+++ b/configuration/src/configs/database.rs
@@ -1,7 +1,7 @@
-use crate::configs::deserialize_data_or_env;
+use near_lake_framework::near_indexer_primitives::near_primitives;
 use validator::Validate;
 
-use near_lake_framework::near_indexer_primitives::near_primitives;
+use crate::configs::deserialize_data_or_env;
 
 // Database connection URL
 // Example: "postgres://user:password@localhost:5432/dbname"
--- a/configuration/src/configs/database.rs
+++ b/configuration/src/configs/database.rs
@@ -1,4 +1,4 @@
-use crate::configs::deserialize_data_or_env;
+use crate::configs::{deserialize_data_or_env, deserialize_url_or_env};
 use near_lake_framework::near_indexer_primitives::near_primitives;
 
 // Database connection URL
@@ -9,7 +9,7 @@ type DatabaseConnectUrl = String;
 pub struct ShardDatabaseConfig {
     #[serde(deserialize_with = "deserialize_data_or_env")]
     pub shard_id: u64,
-    #[serde(deserialize_with = "deserialize_data_or_env")]
+    #[serde(deserialize_with = "deserialize_url_or_env")]
     pub database_url: DatabaseConnectUrl,
 }
 
@@ -35,7 +35,7 @@ impl DatabaseConfig {
 
 #[derive(serde_derive::Deserialize, Debug, Clone, Default)]
 pub struct CommonDatabaseConfig {
-    #[serde(deserialize_with = "deserialize_data_or_env")]
+    #[serde(deserialize_with = "deserialize_url_or_env")]
     pub database_url: DatabaseConnectUrl,
     #[serde(default)]
     pub shards: Vec<ShardDatabaseConfig>,
--- a/configuration/src/configs/general.rs
+++ b/configuration/src/configs/general.rs
@@ -3,7 +3,10 @@ use std::str::FromStr;
 use serde_derive::Deserialize;
 
 use crate::configs::{
-    deserialize_data_or_env, deserialize_optional_data_or_env, required_value_or_panic,
+    deserialize_data_or_env, deserialize_optional_block_cache_size_or_env,
+    deserialize_optional_contract_code_cache_size_or_env, deserialize_optional_data_or_env,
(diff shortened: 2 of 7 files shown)


frolvanya:
@race-of-sloths please, include my PR in the Race

race-of-sloths:
@frolvanya Thank you for your contribution! Your pull request is now a part of the [Race of Sloths](https://race-of-sloths.com)!
New Sloth joined the Race! Welcome!

[<picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://badge.race-of-sloths.com/frolvanya?type=bot&theme=dark">
    <source media="(prefers-color-scheme: light)" srcset="https://badge.race-of-sloths.com/frolvanya?type=bot&theme=light">
    Shows profile picture for the author of the PR
</picture>](https://race-of-sloths.com/profile/frolvanya)


<details><summary>Current status: <i>waiting for scoring</i></summary>

We're waiting for maintainer to score this pull request with `@race-of-sloths score [0,1,2,3,5,8,13]` command
</details><details><summary>What is the Race of Sloths</summary>
<p>
Race of Sloths is a friendly competition where you can participate in challenges and compete with other open-source contributors within your normal workflow

For contributors:

- Tag @race-of-sloths inside your pull requests
- Wait for the maintainer to review and score your pull request
- Check out your position in the [Leaderboard](https://race-of-sloths.com/leaderboard)
- Keep weekly and monthly streaks to reach higher positions
- Boast your contributions with a dynamic picture of your [Profile](https://race-of-sloths.com/profile/frolvanya)

For maintainers:
- Score pull requests that participate in the Race of Sloths
- Engage contributors with fair scoring and fast responses so they keep their streaks
- Promote the Race to the point where the Race starts promoting you
- Grow the community of your contributors

Feel free to check [our website](https://race-of-sloths.com) for additional details!
</p>
</details> 

<details>
<summary>Bot commands</summary>

- For contributors
  - **Include a PR:** `@race-of-sloths include` to enter the Race with your PR
- For maintainers:
  - **Assign points:** `@race-of-sloths score [1/2/3/5/8/13]` to award points based on your assessment.
  - **Reject this PR:** `@race-of-sloths exclude` to send this PR back to the drawing board.
  - **Exclude repo:** `@race-of-sloths pause` to stop bot activity in this repo until `@race-of-sloths unpause` command is called

</details>

kobayurii (changes requested):
I don't think it's the best solution to write new functions for field deserialization and validation every time. Please consider using a third-party crate, for example `serde_valid`.

frolvanya:
> I don't think it's the best solution to write new functions for field deserialization and validation every time. Please consider using a third-party crate, for example `serde_valid`.

Yep, I thought about this and I'm also not a big fan of code duplication. However, I faced few issues, maybe I'm doing something wrong. The `#[validate]` attribute from `serde_valid` is designed to validate fields during serialization and deserialization. However, it is applied directly to fields that are deserialized without custom deserialization functions (`deserialize_with`). When you use `#[serde(deserialize_with)]`, you're bypassing the default deserialization behavior, which includes the `#[validate]` checks. Unfortunately, we can't forget about `deserialize_with` and just stick with normal deserialization + validation workflow, because we still need to check whether we received a value from ENV or from config file

kobayurii:
> > I don't think it's the best solution to write new functions for field deserialization and validation every time. Please consider using a third-party crate, for example `serde_valid`.
> 
> @kobayurii Yep, I thought about this and I'm also not a big fan of code duplication. However, I faced few issues, maybe I'm doing something wrong. The `#[validate]` attribute from `serde_valid` is designed to validate fields during serialization and deserialization. However, it is applied directly to fields that are deserialized without custom deserialization functions (`deserialize_with`). When you use `#[serde(deserialize_with)]`, you're bypassing the default deserialization behavior, which includes the `#[validate]` checks. Unfortunately, we can't forget about `deserialize_with` and just stick with normal deserialization + validation workflow, because we still need to check whether we received a value from ENV or from config file
> 
> P.S: or we need to come up with the custom implementation of validation function and we'll basically end up with the same state of code + a little overhead, because we used an extra crate
> 
> P.S2: + we also need to come up with a custom validation for URLs, because basic `#[validate]` supports only min/max + regex and I don't want to use pure regex to verify a URL

I gave you `erde_valid` as an example. You can choose any other crate that suits us. How about https://docs.rs/validator/latest/validator/. I think it might work for us.

frolvanya:
> > > I don't think it's the best solution to write new functions for field deserialization and validation every time. Please consider using a third-party crate, for example `serde_valid`.
> > 
> > 
> > @kobayurii Yep, I thought about this and I'm also not a big fan of code duplication. However, I faced few issues, maybe I'm doing something wrong. The `#[validate]` attribute from `serde_valid` is designed to validate fields during serialization and deserialization. However, it is applied directly to fields that are deserialized without custom deserialization functions (`deserialize_with`). When you use `#[serde(deserialize_with)]`, you're bypassing the default deserialization behavior, which includes the `#[validate]` checks. Unfortunately, we can't forget about `deserialize_with` and just stick with normal deserialization + validation workflow, because we still need to check whether we received a value from ENV or from config file
> > P.S: or we need to come up with the custom implementation of validation function and we'll basically end up with the same state of code + a little overhead, because we used an extra crate
> > P.S2: + we also need to come up with a custom validation for URLs, because basic `#[validate]` supports only min/max + regex and I don't want to use pure regex to verify a URL
> 
> I gave you `serde_valid` as an example. You can choose any other crate that suits us. How about https://docs.rs/validator/latest/validator/. I think it might work for us.

Oh, I knew only about `serde_valid` and I didn't know about `validator` crate. It might work. Thanks, I'll try to refactor current code

frolvanya:
I rewrote validation to use `validator` crate, but I have few more questions:
1) Is it better to panic if we encounter such config errors or simply send a warn message to logs as we do now? Some of the checks are pretty dangerous (again, shadow_consistency_rate might result to the whole rpc failure), so I'm thinking about throwing a `tracing::error!` along with panic
2) Right now the log message is pretty unreadable and I'm not sure if I should invest some time to make it prettier. The main difficulty is a complexity of `ValidationErrorsKind` structure (it is a enum with multiple different fields). Tell me if this is a good idea to think about the method to improve current log message
image
image

P.S: @kobayurii Thanks for sharing such powerful crate with me! It might become handy in my pet projects

khorolets:
@race-of-sloths score 2

@frolvanya 
> 1. Is it better to panic if we encounter such config errors or simply send a warn message to logs as we do now?

In such dangerous situations, I'd prefer to panic.

As for the second part of your question, let's keep "pretty unreadable" for now. We'll get back to it if we realize it's a real issue for us.

configuration/src/configs/database.rs
```diff
@@ -1,14 +1,17 @@
 use crate::configs::deserialize_data_or_env;
+use validator::Validate;
```
khorolets, line 2: While you're here, please fix the order of imports:

- std
- 3rd-party
- project's crates
- modules

race-of-sloths:
### 🌟 Score recorded!

@khorolets, thank you for scoring this pull request in the Race of Sloths!

configuration/src/configs/general.rs
```diff
@@ -95,16 +101,23 @@ impl FromStr for ChainId {
...
 }
 
-#[derive(Deserialize, Debug, Clone)]
+#[derive(Validate, Deserialize, Debug, Clone)]
 pub struct CommonGeneralRpcServerConfig {
     #[serde(deserialize_with = "deserialize_optional_data_or_env", default)]
     pub server_port: Option<u16>,
     #[serde(deserialize_with = "deserialize_optional_data_or_env", default)]
     pub max_gas_burnt: Option<u64>,
+    #[validate(range(min = 0.0, message = "Contract code cache size must be greater than 0"))]
```
khorolets, line 110: Would `min = 0.0` allow that value? I mean the error message clearly says it should be greater than 0 and I suspect that `min` can allow it

khorolets (approved):
Great job, I left a few minor comments and approve 👍

configuration/src/lib.rs
```diff
@@ -18,6 +20,11 @@ where
     let path_root = find_configs_root().await?;
     load_env(path_root.clone()).await?;
     let common_config = read_toml_file(path_root).await?;
+
+    if let Err(validation_errors) = common_config.validate() {
```
kobayurii, line 24: I expect configuration validation to panic and prevent the service from running with a non-valid configuration.
It is better to see a panic and error message than to see a warning that may just get lost in a large number of logs.

kobayurii (approved):
Great! I left a one comment and approve 👍

frolvanya:
I've fixed everything. As I see both of you approved the PR, so I'll merge it

race-of-sloths:
### ✅ PR is finalized! 
Your contribution is much appreciated with a final score of 2!
You have received 30 (20 base + 10 weekly bonus) Sloth points for this contribution


Another weekly streak completed, well done @frolvanya! To keep your weekly streak and get another bonus make pull request next week! Looking forward to see you in race-of-sloths

PR #353, opened 2024-09-19
(rpc_server): allow to fetch state key by prefix without state_size_limit

kobayurii:
In this pull request, we have added the ability to always request a state key by prefix without state_size_limit.


khorolets (approved):
@race-of-sloths score 2

PR #355, opened 2024-10-02
migrate to fastnear

kobayurii:
This PR brings significant improvements to the read_rpc server after adding `fastnear` as an alternative data provider for NEAR Lake (https://github.com/near/near-lake-framework-rs/pull/109).

**Improvements to read_rpc Server:**

- Leveraged the introduction of fastnear to optimize the read_rpc server.
- Simplified data processing logic of final and optimistic blocks, reducing latency and increasing throughput.
- Improved system to fetch blocks and chunks to enhance reliability and robustness, particularly in high-load scenarios.

These enhancements contribute to better overall performance of the read_rpc server, resulting in higher reliability, faster response times, and a more resilient system.


kobayurii:
@race-of-sloths please, include my PR in the Race

rpc-server/src/config.rs
```diff
@@ -20,8 +20,7 @@ pub struct GenesisInfo {
 impl GenesisInfo {
     pub async fn get(
         near_rpc_client: &crate::utils::JsonRpcClient,
-        s3_client: &near_lake_framework::s3_fetchers::LakeS3Client,
-        s3_bucket_name: &str,
+        fastnear_client: &near_lake_framework::fastnear_client::FastNearClient,
```
khorolets, line 23: This change basically indicates disadvantages we are about to introduce to the `near-lake-framework`. I would expect `near_lake_framework::Client` here that could work with either Lake data or FastNear data. We need to rethink the changes on the framework. Though, it is not a blocker for this PR

rpc-server/src/utils.rs
```diff
@@ -146,20 +145,27 @@ pub async fn get_current_validators(
     Ok(near_rpc_client.call(params, None).await?)
 }
 
+pub async fn get_current_protocol_version(
```
khorolets, line 148: btw could you introduce the metric with protocol version to expose, just in case

rpc-server/src/utils.rs
```diff
@@ -178,142 +184,62 @@ async fn handle_streamer_message(
...
-                .await
-                {
-                    tracing::error!("Error to handle_streamer_message: {:?}", err);
-                }
-            }
-            Err(err) => {
-                tracing::warn!("Error to get final block from redis: {:?}", err);
-            }
-        }
+            tracing::error!("Error to handle_streamer_message: {:?}", err);
```
khorolets, line 208:
```suggestion
            tracing::error!("Error in fn handle_streamer_message(): {:?}", err);
```

khorolets (approved):
Looks great to me, I've highlighted a few places to note. However, nothing critical so far, so we're good to merge even without addressing the things I've mentioned in the PR review. Great job!

rpc-server/src/config.rs
kobayurii: I will improve it on the near-lake-framework side.

frol:
Hey, unfortunately, we skipped this PR as our bot was blocked  :( But we extended the time range so should pass just now. Sorry for the inconvenience.

@race-of-sloths include

PR #394, opened 2025-03-06
New database schema

kobayurii:
We faced the problem that when requesting state changes, requests run for more than 2 minutes, which is unacceptable. To optimize this, we developed and introduced a new method of data storage.

In this PR, we implement a new database schema.

New method used to track historical changes in a database while maintaining the ability to query the latest state efficiently. It works by storing each change as a new row and marking old values as expired.


kobayurii:
With this schema, it is impossible to indexing data.

